PR #10, opened 2022-06-07
Added metadata endpoint

### LINKED ISSUE #2: Add BOM Meta Endpoint
The idea behind a BOM Meta endpoint is to provide format, hash, and external signature information.

The BOM Meta retrieval would work similar to the existing BOM retrieval, but would return metadata rather than the BOM itself. 

```
bom-meta-url = system-url "/" bom-identifier
bom-identifier    = segment
                        ; an identifier that uniquely identifies a BOM
                        ; NOTE: MUST be appropriately URI encoded
                        ; segment as defined in RFC3986
```

As an example, here's a snippet response for what I'm thinking about being returned:

```json
{
  "spec": {
    "format": "CycloneDX",
    "version": "1.4",
    "mime-type": "application/vnd.cyclonedx+json",
  },
  "published": "2022-03-07T15:50+00Z",
  "checksum": [
    { "alg": "SHA-256", "value": "CF80CD8..." },
    { "alg": "SHA-512", "value": "CF80CD8..." },
    { "alg": "SHA3-256", "value": "CF80CD8..." },
    { "alg": "SHA3-512", "value": "CF80CD8..." },
    { "alg": "BLAKE3", "value": "CF80CD8..." }
  ],
  "signature": [
  
  ]
}
```

I think `alg` should be an enum with only those supported algorithms.

As for signatures, it would be ideal if we could support external signature files, signature services (e.g sigstore), and external inline.

stevespringett:
Closes #2


README.md
```diff
@@ -147,3 +147,123 @@ Clients MUST provide the correct `Content-Type` HTTP request header. For example
...
+
+The server MUST respond with a JSON payload using the following syntax:
+
+```json
+{
+  "identifier": "",
+  "spec": {
+    "format": "",
+    "version": ""
+  },
```
coderpatros, line 188: Isn't spec unnecessary? The BOM may be available in multiple formats. And even if it's only available in one format, that also needs to be specified in the artifacts array.

README.md
```diff
@@ -147,3 +147,123 @@ Clients MUST provide the correct `Content-Type` HTTP request header. For example
...
+If no artifacts are available, the "artifacts" object MUST be an empty array.
+
+##### The MIME-TYPES property
+The "mime-type" property MUST be a string that conform to [RFC-2045](https://datatracker.ietf.org/doc/html/rfc2045).
+This property specifies the media type the underlying server maintains which can be requested.
+
+##### The CHECKSUM property
+The "checksum" property describes an array of objects, each MUST contain "alg" and "value". The "alg" property
+is a REQUIRED string and must be one of SHA-1, SHA-256, SHA-384, SHA-512, SHA3-256, SHA3-384, SHA3-512, BLAKE2b, or BLAKE3.
+The "value" property is a hexadecimal string derived from the hashing function. All values MUST be uppercase.
```
coderpatros, line 225: I wonder if we should reference the CycloneDX spec for the available checksum values. That way we don't need to update this every time we add new algorithms to the spec.

README.md
```diff
@@ -147,3 +147,123 @@ Clients MUST provide the correct `Content-Type` HTTP request header. For example
...
+    {
+      "mime-type": "",
+      "checksum": [
+        { "alg":  "", "value":  ""}
+      ]
+    }
+  ],
+  "published": ""
+}
+```
```
coderpatros, line 199: Should we define this with a JSON schema instead? That might be a bit tricky with my other comment regarding hash algorithms referencing CDX spec though.

oej:
This will have to be rebased as the spec it changes moved to a separate document

oej:
Do we want to go ahead and merge this or delete the pull request. Ping @stevespringett

oej:
Closing this pull request.

PR #14, opened 2024-05-31
Update documentation from our Google doc "use cases" documentation

oej:
- update readme based on the new work
- add information about ECMA
- move the existing API to separate document


oej:
Added use cases and requirements

PR #21, opened 2024-07-13
Adding a first draft on handling of digital signatures

oej:
Trying to gather thoughts on digital signatures in the TEA collections and for TEA artefacts.

Feedback is always welcome!

### DIFF
--- a/signatures/signature.md
+++ b/signatures/signature.md
@@ -44,7 +44,7 @@ best current practise.
 ## API trust
 
 The TEA API is built on the HTTP protocol with TLS encryption
-and authentication, using the https:// URL scheme.
+and authentication, using the `https://` URL scheme.
 
 The TLS server certificate is normally issued by a public Certificate
 Authority that is part of the Web PKI. The client needs to validate
--- a/signatures/signature.md
+++ b/signatures/signature.md
@@ -8,15 +8,14 @@ signature files.
 
 - __Integrity__: Documents dowloaded needs to be the same
   as documents published
-- __Identity__: 
-  - Customers needs to be able to verify the 
+- __Identity__:
+  - Customers need to be able to verify the
   publisher of the documents and verify that it is
   the expected publisher.
   - A TEA server may want to verify that published
   documents are signed by the expected publisher
   and that signatures are valid.
   
-
 In order to sign, a pair of asymmetric keys will be needed.
 The public key is used to create a certificate, signed
 by a certificate authority (CA).
@@ -27,7 +26,7 @@ parties do not automatically trust that internal PKI.
 
 This document outlines a proposal on how to build that trust and
 make it possible for publishers to use an internal PKI. It is
-of course important that this PKI is maintained according to 
+of course important that this PKI is maintained according to
 best current practise.
 
 ## API trust
@@ -59,13 +58,13 @@ that would cause a chicken-and-egg problem.
 ## Digital signatures
 
 ### Digital signatures as specified for CycloneDX
+
 "Digital signatures may be applied to a BOM or to an assembly within a BOM.
 CycloneDX supports XML Signature, JSON Web Signature (JWS), and JSON Signature Format (JSF).
 Signed BOMs benefit by providing advanced integrity and non-repudiation capabilities."
-https://cyclonedx.org/use-cases/#authenticity
+<https://cyclonedx.org/use-cases/#authenticity>
 
-
-### External (deattached) digital signatures for other documents
+### External (detached) digital signatures for other documents
 
 - indication of hash algorithm
 - indicator of cert used
@@ -76,7 +75,7 @@ https://cyclonedx.org/use-cases/#authenticity
 ## Using Sigstore for signing
 
 Sigstore is an excellent free service for both signing of GIT commits as well
-as artefacts by using ephemeral certificates (very shortlived) and a 
+as artefacts by using ephemeral certificates (very shortlived) and a
 certificate transparency log for validation and verification.
 Sigstore signatures contain timestamps from a timestamping service.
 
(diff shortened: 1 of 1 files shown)


signatures/signature.md
```diff
@@ -0,0 +1,108 @@
+# Trusting digital signatures in TEA
+
+Software transparency is a lot about trust. Within the
```
aj-stein-nist, line 3: Trust is a human determination about software. Is software transparency not about trustworthiness?

signatures/signature.md
```diff
@@ -0,0 +1,108 @@
...
+The TEA API has a `/trust-anchors/` API that will download
+the current trust anchor APIs. This file is not signed,
+that would cause a chicken-and-egg problem.
+
+## Validating the trust anchors using DNSsec (DANE)
+
+## Digital signatures
+
+### Digital signatures as specified for CycloneDX
+"Digital signatures may be applied to a BOM or to an assembly within a BOM.
```
aj-stein-nist, line 62: Is the intention here that you would have digital signatures for individual inline assemblies and the overall package that composes them? I tried reviewing the CycloneDX documents and this specification but that sounds like it can get complex very quickly.

signatures/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+- indication of hash algorithm
+- indicator of cert used
+- intermediate cert and sign cert
+
+### Validating the digital signature
+
+## Using Sigstore for signing
+
+Sigstore is an excellent free service for both signing of GIT commits as well
```
aj-stein-nist, line 78: Is it the responsibility of the Transparency API described in this specification to compute and verify such hashes from external sources like Sigstore and/or elsewhere?

ppkarwasz:
I like the way this proposal aims to automate the verification of signatures of published artifacts. We regularly receive requests from users that wish to verify the artifact signatures we publish on Maven Central.

I have however a couple of concerns:

* CycloneDX signatures are a new mechanism to sign artifacts and currently most SBOMs don't have them. On the other hand I found GPG signatures on Maven Central going back as far as 2007 at least. We should probably also publish trust anchors for these legacy mechanisms, which would allow us to verify the entire software stack of an application without waiting for an SBOM signature of each dependency.
* Should the API distribute only its own trust anchors or also the trust anchors of its suppliers (using some namespacing mechanism? The latter seems like a better choice, since it will create a web-of-trust that is more difficult to compromise.

oej:
Supporting GPG signatures is good for backwards compatibility but do we want to put that implementation burden on new implementations, which this will be? Food for thought.

Being able to add upstream trust anchors is a very good idea.

ppkarwasz:
> Supporting GPG signatures is good for backwards compatibility but do we want to put that implementation burden on new implementations, which this will be? Food for thought.

The availability of PGP trust anchors is the main reason PyPI removed signatures from their repository: https://blog.pypi.org/posts/2023-05-23-removing-pgp/
On the other hand Maven Central requires PGP signatures for all published artifacts: https://central.sonatype.org/publish/requirements/

I don't think we should require the support of the OpenPGP format, but we might allow users to publish the public keys as X509 certificates and mark their usage as "Maven only", "PyPI only", etc.
The main use case would be to allow a user to verify the authenticity of the sub-components of an _ACME Miracle Server_ product. If Dependency Track can verify that a sub-component comes from an upstream supplier **without modifications**, it can provide security alerts **before** those alerts propagate to the _ACME_ API server.

signatures/signature.md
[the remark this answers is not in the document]
oej: I will rephrase this... Thanks for the feedback.

signatures/signature.md
```diff
@@ -0,0 +1,108 @@
...
+
+In order to sign, a pair of asymmetric keys will be needed.
+The public key is used to create a certificate, signed
+by a certificate authority (CA).
+
+A software publisher may buy CA services from a commercial vendors
+or set up an internal solution. The issue with that is that external
+parties do not automatically trust that internal PKI.
+
+This document outlines a proposal on how to build that trust and
```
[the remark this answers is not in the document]
oej: I think signing has to be optional. But the question of a third party that publishes something is a bit out of scope for our work, while I understand it is part of SCITT. I will consider this and see what can be done.

signatures/signature.md
oej: I just quoted the text from the CycloneDX specs. Agree that it may feel messy, but it may have a point as you can include attestations, patches and much more in a CycloneDX BOM.

signatures/signature.md
oej: That is a good question. I think validation primarily is when someone downloads artefacts. But it can of course also apply to publication - a server implementation may not accept signed documents that the *server* can not validate. Good question.

oej:
Moved discussions on PGP to a separate issue

signatures/signature.md
oej: Personally I do like detached signatures :-)

oej:
Added a clarification that digital signatures are optional. It may not be in future versions of the API, but we have to start somewhere.

signatures/signature.md
```diff
@@ -0,0 +1,122 @@
...
+Within the
+API documents can be signed with an electronic
+signature. CycloneDX boms supports signatures within
+the JSON file, but other artefacts may need external
+signature files, a detached signature.
+
+- __Integrity__: Documents dowloaded needs to be the same
+  as documents published
+- __Identity__:
+  - Customers need to be able to verify the
```
madpah, line 20:
```suggestion
  - Customers need to be able to verify the identity of the
```

signatures/signature.md
```diff
@@ -0,0 +1,122 @@
...
+
+This document outlines a proposal on how to build that trust and
+make it possible for publishers to use an internal PKI. It is
+of course important that this PKI is maintained according to
+best current practise.
+
+## API trust
+
+The TEA API is built on the HTTP protocol with TLS encryption
+and authentication, using the https:// URL scheme.
```
madpah, line 44:
```suggestion
and authentication, using the `https://` URL scheme.
```

signatures/signature.md
```diff
@@ -0,0 +1,122 @@
...
+level of validation.
+
+## Getting trust anchors
+
+Much like the EST protocol, the TEA protocol can be used
+to download trust anchors for a private PKI. These are
+PEM-encoded certificates in one single text file.
+
+The TEA API has a `/trust-anchors/` API that will download
+the current trust anchor APIs. This file is not signed,
```
madpah, line 65: `the current trust anchor APIs. This file is not signed,`

Is APIs the correct word here @oej ?

signatures/signature.md
jksolbakken: With OCI (Docker) images the producer generates and uploads the signature to the registry alongside the image artifact using the `cosign` utility. The consumer also uses cosign to [verify](https://docs.sigstore.dev/cosign/verifying/verify/) the signature against the Sigstore transparency log.

signatures/signature.md
```diff
@@ -0,0 +1,122 @@
...
+that would cause a chicken-and-egg problem. The certificates
```
ppkarwasz, line 66: > This file is not signed, that would cause a chicken-and-egg problem.

Should we exclude _à priori_ the signing of trust anchors? In the future they might be signed by some authority, e.g. regulators.

PR #22, opened 2024-07-30
tea collection: Adding a first description

oej:
Adding an outline for discussion in our meeting.

### DIFF
--- a/tea-collection/tea-collection.md
+++ b/tea-collection/tea-collection.md
@@ -7,6 +7,14 @@ TEA collections.
 The TEA collection is normally created by the TEA application server at
 publication time of artifacts.
 
+If there are any updates of artifacts within a collection for the same
+version of a product, then a new TEA Collection object is created and signed.
+This update will have the same UUID, but a new version number. A reason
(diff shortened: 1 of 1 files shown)


tea-collection/tea-collection.md
```diff
@@ -0,0 +1,48 @@
+# The TEA Collection object
+
+For each product and version there is a Tea Collection object, which is a list
```
aj-stein-nist, line 3: Is the list mutable and have a specific ordering? (Time of insertion in relation to others collection items in the list? Another ordering method?)

oej: That is a good question. Let me take a stab and propose something. In my head, the collection is a signed JSON file. When updating, a new version of the collection file is produced.

oej: Added a commit clarifying that order is insignificant. Earlier on I added a timestamp for publishing the collection in the main object.

This may be wrong, but is at least an answer. Did you have a use case in mind that would require a specific ordering?

oej:
WIll merge this as a proposal for now.

Issue #67, opened 2024-11-13
Using security.txt to indicate TEA service availability


oej:
> As an alternative, we could register a [`security.txt`](https://securitytxt.org/) field.
> 
> The list of `security.txt` fields is also [a registry maintained by IANA](https://www.iana.org/assignments/security-txt-fields/security-txt-fields.xhtml). 

 _Originally posted by @ppkarwasz in [#30](https://github.com/CycloneDX/transparency-exchange-api/issues/30#issuecomment-2414885101)_

oej:
Security.txt is a good way to be able to find the API without having to have a product ID. It's different than the "ordinary" TEA discovery based on the TEI.

oej:
From the RFC: "Designated experts should determine whether a proposed registration or update provides value to organizations and researchers using this format and makes sense in the context of industry-accepted vulnerability disclosure processes such as [[ISO.29147.2018](https://www.rfc-editor.org/rfc/rfc9116#ISO.29147.2018)] and [[CERT.CVD](https://www.rfc-editor.org/rfc/rfc9116#CERT.CVD)]."

Issue #66, opened 2024-11-13
TEA Discovery: Should we be opinionated on priority of different discovery mechanisms?


madpah:
For discussion.

The current in-draft [Discovery](https://github.com/CycloneDX/transparency-exchange-api/tree/main/discovery) docs cater for **TWO** mechanisms to discovery the TEA API for a given TEI:

1. Discovery using DNS (subject to #64 - will be SVCB records)
2. Discovery using Host (use of AAAA or A records) to get to a WebServer at the Host so that `/.well-known/tei` can be accessed - this would likely produce a HTTP Redirect (301 or 302 - TBC which are permitted within TEI specification) to the TEA API

Additionally, @ppkarwasz has suggested (in https://github.com/CycloneDX/transparency-exchange-api/issues/30 - now tracked in https://github.com/CycloneDX/transparency-exchange-api/issues/67):

3. Discovery using Host (use of AAAA or A records) to get to a WebServer at the Host so that `/.well-known/security.txt` can be accessed, containing a new (yet to proposed and [registered](https://www.iana.org/assignments/security-txt-fields/security-txt-fields.xhtml)) field that provides the URL to the TEA API

When considering implementation of the TEA Specification - I would suggest it is prudent for the Specification to be opinionated on which method(s) have priority along with reasons.

FYI @oej

oej:
I think like this:
For TEI discovery:
DNS is the best way to(provides failover etc). If that's not allowed, #2 will be used.

The security.txt case is very different as it a way to say "by the way, we have a TEA service". For that to work, there needs to be a way to get product identifiers. It requires further thinking.

oej:
https://github.com/CycloneDX/transparency-exchange-api/pull/72

Issue #30, opened 2024-08-20
Register .well-known URI for TEA


oej:
The IANA maintains a registry of .well-known URI's

ppkarwasz:
As an alternative, we could register a [`security.txt`](https://securitytxt.org/) field.

The list of `security.txt` fields is also [a registry maintained by IANA](https://www.iana.org/assignments/security-txt-fields/security-txt-fields.xhtml).

madpah:
This seems like a sane thing to do - inline with https://github.com/CycloneDX/transparency-exchange-api/tree/main/discovery#tei-resolution-using-dns.

oej:
I think we need a separate issue on Security.txt, because it's such a different approach. The other discovery starts with "I have a product identifier.

The security.txt case is very different as it a way to say "by the way, we have a TEA service". For that to work, there needs to be a way to get product identifiers. It requires further thinking.

oej:
Created #67 for security.txt so this issue is focused on .well-known

oej:
So, I don't want to have an API running on the .well-known URI. Should we define a syntax, inspired by security.txt for redirection to another URL?

`TEA-baseurl: https://product.example.com/tea/v1/`

I don't think simply running a 302 is a good architecture for this. It requires less parsing though.

madpah:
For `security.txt` agree we need to define the complete URL to the TEA API Service.

For `/.well-known/tei` I'd have thought an appropriate HTTP 30x redirect to the TEA API Service would be good enough?

oej:
I think the API is "tea" so let's go for tea. I don't know if there's a best current practice for applying a 302, need to check that. Feels raw, but it will work.

oej:
https://www.rfc-editor.org/rfc/rfc8615

I don't find any text against 302 redirects. Good.

ppkarwasz:
> https://www.rfc-editor.org/rfc/rfc8615
> 
> I don't find any text against 302 redirects. Good.

Some well known URIs even require a 302 redirect. E.g. [RFC 6764 (calDAV/cardDAV)](https://www.rfc-editor.org/rfc/rfc6764.html#page-4) states:

> Two ".well-known" URIs are registered by this specification for  CalDAV and CardDAV services, "caldav" and "carddav" respectively …
> These URIs point to a resource that the client can use as the initial "context path" for the service they are trying to connect to.
[...]

PR #68, opened 2024-11-13
Updated Discovery to cater for move to SVCB DNS records

### LINKED ISSUE #64: Which DNS Record Type to use in Discovery? `URI` Record Type is not widely supported
https://github.com/CycloneDX/transparency-exchange-api/blob/main/discovery/readme.md#tei-resolution-using-dns currently states that DNS Records of type `URI` will be used to resolve the list of Transparency Exchange endpoints for the TEI in question.

However, the DNS record type `URI` does not seem to be supported widely - see AWS Route 53 screenshot where it is not available:
Image

Should we consider using another DNS record type - such as `TXT` that is more widely supported?

FYI @oej @stevespringett

madpah:
This PR primarily drafts changes to resolve #64.

Additional clarifications have been made in relation to other work under discussion:
- #30 
- #66 
- #67

### DIFF
--- a/discovery/readme.md
+++ b/discovery/readme.md
@@ -34,20 +34,19 @@ to global uniqueness without new registries.
 The TEI can be shown in the software itself, in shipping documentation, in web pages and app stores.
 TEI is unique for a product, not a version of a software. The TEI consist of three core parts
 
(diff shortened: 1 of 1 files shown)


oej (approved):
Thank you!

PR #69, opened 2024-11-14
beginning of draft OpenApi Spec for TEA

madpah:
Only considering Consumer of TEA at this time.

Sharing as a draft to aid discussions.

Schema can be visualised by using https://editor-next.swagger.io/

### DIFF
--- a/spec
+++ b/spec
@@ -1,0 +1,405 @@
+{
+    "$schema": "https://spec.openapis.org/oas/3.1/dialect/base",
+    "openapi": "3.1.1",
+    "info": {
+        "title": "Transparency Exchange API",
+        "summary": "Transparency Exchange API specification for consumers and publishers",
+        "description": "TBC",
+        "contact": {
+            "name": "TEA Working Group",
+            "email": "TBC",
+            "url": "https://github.com/CycloneDX/transparency-exchange-api"
+        },
+        "license": {
+            "identifier": "Apache-2.0",
+            "url": "https://github.com/CycloneDX/transparency-exchange-api/blob/main/LICENSE"
+        },
+        "version": "0.0.1"
+    },
+    "servers": [
+        {
+            "url": "http://localhost/tea/v1",
+            "description": "Local development"
+        }
+    ],
+    "paths": {
+        "/index/{product-identifier}": {
+            "get": {
+                "description": "Access the TEA Index for the supplied product identifier",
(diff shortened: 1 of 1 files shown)


oej:
I am sorry I did not tell you, but there's earlier work here: https://github.com/oej/pavel-project-koala

I think we can use your code as a starting point.

At a later point, we may want to add a validator and check with generators as welll in CI/CD

spec
```diff
@@ -0,0 +1,405 @@
...
+    "servers": [
+        {
+            "url": "http://localhost/tea/v1",
+            "description": "Local development"
+        }
+    ],
+    "paths": {
+        "/index/{product-identifier}": {
+            "get": {
+                "description": "Access the TEA Index for the supplied product identifier",
```
oej, line 28: In the docs we call it the "TEA product index (TPI)"

madpah:
> I am sorry I did not tell you, but there's earlier work here: https://github.com/oej/pavel-project-koala

Will check it out.

> I think we can use your code as a starting point.
> 
> At a later point, we may want to add a validator and check with generators as well in CI/CD

Happy to add generators and validators here.

spec
madpah: Given, in this example, the whole API is under `/tea/v1`, the context of being under `tea` is a given. So we can rename from `index` to `product-index` to align with the docs @oej ?

oej: That may be a good thing. Coming back to earlier discussions about security.txt discovery - do we need to have a paginated list at /tea/product-index/ for those that doesn't start with a TEI ?

madpah: Aligned with docs - now at `/product-index`.

spec
```diff
@@ -0,0 +1,405 @@
...
+                "tags": [
+                    "TEA Leaf"
+                ]
+            }
+        }
+    },
+    "webhooks": {},
+    "components": {
+        "schemas": {
+            "tea-index": {
```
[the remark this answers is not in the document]
madpah: I saw above - updated.

madpah:
@oej - Added OpenAPI Client generation for Go, Java Webclient, Python and Typescript.

This does not publish - this is just for PR validation etc. CI-style.

- Validates OpenAPI Spec
- Proves libraries generate
- Run any tests per library

oej (approved):
Great work! Thank you!

PR #72, opened 2024-11-18
Discovery: Remove load balancing and failover support using DNS

oej:
- Remove the DNS URI records as a few large DNS providers does not support them (and the RFC is informational)
- Clarify TLS certificate requirements (according to RFC 9640)
- Fix some markdown formatting (make markdownlint happy)

### DIFF
--- a/discovery/readme.md
+++ b/discovery/readme.md
@@ -174,7 +174,8 @@ in DNS. The URI is composed of the name with the  "/.well-known/tea" prefix adde
 tea.example.com.   3600 IN HTTPS  1  tea01.prod.example.com.
 ```
 
-Results in `https://tea.example.com/.well-known/tea/<identifier>` while connecting to
+Results in the base URI (without the product identifier) 
+`https://tea.example.com/.well-known/tea/` while connecting to
 the host tea01.prod.example.com.
 
 ### Load balancing and fail over
--- a/discovery/readme.md
+++ b/discovery/readme.md
@@ -170,7 +170,7 @@ After selecting a host and resolving that to IP address, the TEA client
 connects to the host using HTTPS with the original name, not the one found
 in DNS. The URI is composed of the name with the `/.well-known/tea` prefix added.
 
-```text
+```zone_file
 tea.example.com.   3600 IN HTTPS  1  tea01.prod.example.com.
 ```
 
--- a/discovery/readme.md
+++ b/discovery/readme.md
@@ -192,7 +192,19 @@ If they are not reachable, tea03 will be used for failover.
 
 It is recommended to have a third party external repository as the last priority.
 
-### Finding the Index using DNS result
+## Connecting to the API
+
+When connecting to the ".well-known/tea" URI with the unique identifier
+a HTTP redirect is **required**.
+
+The server MUST redirect HTTP requests for that resource
+to the actual "context path" using one of the available mechanisms
+provided by HTTP (e.g., using a 301, 303, or 307 response).  Clients
+MUST handle HTTP redirects on the ".well-known" URI.  Servers MUST
+NOT locate the actual TEA service endpoint at the
+".well-known" URI as per Section 1.1 of [RFC5785].
+
+### Overview: Finding the Index using DNS result
 
 Append the product part of the TEI to the URI found
 
--- a/discovery/readme.md
+++ b/discovery/readme.md
@@ -37,7 +37,7 @@ The TEI for a product can be communicated to the user in many ways.
 
 ## TEA Discovery - defining an extensible identifier
 
-TEA discovery is the process where a user with a product identifier can discover and downloadg
+TEA discovery is the process where a user with a product identifier can discover and download
 artifacts automatically, with or without authentication. A globally unique identifier is
 required for a given product. This identifier is called the Transparency Exchange Identifier (TEI).
 
@@ -65,7 +65,7 @@ urn:tei:<type>:<domain-name>:<unique-identifier>
 ````
 
 - The **`type`** which defines the syntax of the unique identifier part
-- The **`domain-name`** part does not have to exist as a web server (HTTPS), but may do
+- The **`domain-name`** part resolves into a web server, which may not be the API host.
   - The uniqueness of the name is the domain name part that has to be registred at creation of the TEI.
 - The **`unique-identifier`** has to be unique within the `domain-name`.
   Recommendation is to use a UUID but it can be an existing article code too
@@ -89,6 +89,7 @@ urn:tei:purl:<domain-name>:<purl>
 ````
 
 Example:
+
 ```text
 urn:tei:purl:cyclonedx.org:pkg:pypi/cyclonedx-python-lib@8.4.0?extension=whl&qualifier=py3-none-any
 ```
@@ -109,9 +110,9 @@ Note that there is a TEI SWID type as well as a PURL SWID type.
 
 Where the `unique-identifier` is a Hash. Supports the following hash types:
 
-* SHA256
-* SHA384
-* SHA512
+- SHA256
+- SHA384
+- SHA512
 
 ```text
 urn:tei:hash:<domain-name>:<hashtype>:<hash>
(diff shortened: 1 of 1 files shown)


discovery/readme.md
```diff
@@ -128,58 +142,76 @@ Has to be a valid UUID.
...
+in DNS. The URI is composed of the name with the  "/.well-known/tea" prefix added.
 
+```text
+tea.example.com.   3600 IN HTTPS  1  tea01.prod.example.com.
 ```
-_tei._tcp.tex.example.com.   3600 IN URI 10 1 “https://www.example.com/transparency“
-_tei._tcp.tex.example.com.   3600 IN URI 20 1 “https://backup.example.com/transparency“
-_tei._tcp.tex.example.com.   3600 IN URI 30 1 “https://thirdparty.example.org/example.com/transparency“
+
+Results in `https://tea.example.com/.well-known/tea/<identifier>` while connecting to
```
madpah, line 173: I'm unclear why we'd connect to a host that is different from the URL - doesn't this just mean we'd connect to `https://tea01.prod.example.com/.well-known/tea/<identifier>` ?

madpah: Also, I'd drop the `<identifier>` from this example as that is subject to the API Specification

discovery/readme.md
```diff
@@ -128,58 +142,76 @@ Has to be a valid UUID.
...
 - TEI: `urn:tei:uuid:products.example.com:d4d9f54a-abcf-11ee-ac79-1a52914d44b1`
-- DNS record: `_tei._tcp.products.example.com`
-- URI in DNS: `://www.example.com/transparency/`
-- URL: `https://www.example.com/transparency/d4d9f54a-abcf-11ee-ac79-1a52914d44b1/`
+- DNS record: `products.example.com`
+- Server found in DNS HTTPS record: `teapot.prod.example.com`
+- URL: `https://products.example.com/.well-known/tea/d4d9f54a-abcf-11ee-ac79-1a52914d44b1/`
+- HTTP 302 redirect to "https://teapot02.consumer.example.com/tea/v2/product-index/d4d9f54a-abcf-11ee-ac79-1a52914d44b1'
+
+The server at "teapot.prod.example.com" needs a TLS certificate including "products.example.com"
```
madpah, line 201: Why would we not just connect to `teapot.prod.example.com`? I don't understand the value provided by the TLS cert at `teapot.prod.example.com` also including a Subject Alt Name of `products.example.com`

discovery/readme.md
```diff
@@ -128,58 +142,76 @@ Has to be a valid UUID.
...
 
 Append the product part of the TEI to the URI found
 
 - TEI: `urn:tei:uuid:products.example.com:d4d9f54a-abcf-11ee-ac79-1a52914d44b1`
-- DNS record: `_tei._tcp.products.example.com`
-- URI in DNS: `://www.example.com/transparency/`
-- URL: `https://www.example.com/transparency/d4d9f54a-abcf-11ee-ac79-1a52914d44b1/`
+- DNS record: `products.example.com`
+- Server found in DNS HTTPS record: `teapot.prod.example.com`
+- URL: `https://products.example.com/.well-known/tea/d4d9f54a-abcf-11ee-ac79-1a52914d44b1/`
```
madpah, line 198: As per above - I'm unclear why the URL would not be `https://teapot.prod.example.com/.well-known/tea/d4d9f54a-abcf-11ee-ac79-1a52914d44b1/`

madpah (commented):
Comments add @oej

discovery/readme.md
oej: DNS is just used to find the server for the HTTPS://products.example.com URI 

I parse the RFC as this is NOT a redirection which would change the URI.

oej: I don't know if any web browsers or curl actually use these records. Will check with CURL.

oej:
Text from the RFC:  "Note that none of these forms alter the origin or authority for
   validation purposes.  For example, TLS clients MUST continue to
   validate TLS certificates for the original service name."

So if the URI points to "products.example.com" - this is what you use to validate the TLS cert for the *first* server. After 302 you got a new or the same server.

discovery/readme.md
oej: Because the "user" is connecting to "products.example.com" - all the other host names just show up in DNS resolution to find a proper HTTP server to connect to, but they don't change the URI that was requested.

discovery/readme.md
oej: Fixed in 7396851

discovery/readme.md
madpah: I guess this is akin to using `CNAME` records with regular HTTP web browsing...

oej:
https://www.netmeister.org/blog/https-rrs.html

“Firefox has been making HTTPS lookups (albeit only over DoH) [since May 2020](https://bugzilla.mozilla.org/show_bug.cgi?id=1623126), Apple’s iOS and Safari / macOS [since September 2020](https://mailarchive.ietf.org/arch/msg/quic/sFgifP9vOY9xsmogVqiq-qtxPiQ/), Chrome has had partial support [since December 2020](https://groups.google.com/a/chromium.org/g/blink-dev/c/brZTXr6-2PU), and just recently[enabled ECH](https://groups.google.com/a/chromium.org/g/blink-dev/c/CmlXjQeNWDI/m/hx-_4lNBAQAJ) by default. Various [DNS](https://ns1.com/blog/ns1-announces-support-for-svcb-and-https-records) [service providers](https://community.akamai.com/customers/s/article/NetworkOperatorCommunityNewSVCBHTTPSResourceRecordsinthewild20201128135350?language=en_US) also offer support for HTTPS and SVCB records already.”

discovery/readme.md
oej: As CNAME can't be used at top level, like "https://edvina.net" these records will work there. Yes, it's similar.

discovery/readme.md
```diff
@@ -128,58 +142,89 @@ Has to be a valid UUID.
...
-### Finding the Index using DNS result
+## Connecting to the API
+
+When connecting to the ".well-known/tea" URI with the unique identifier
+a HTTP redirect is **required**.
+
+The server MUST redirect HTTP requests for that resource
+to the actual "context path" using one of the available mechanisms
+provided by HTTP (e.g., using a 301, 303, or 307 response).  Clients
+MUST handle HTTP redirects on the ".well-known" URI.  Servers MUST
```
madpah, line 200:
```suggestion
MUST handle HTTP redirects on the `.well-known` URI.  Servers MUST
```

discovery/readme.md
```diff
@@ -128,58 +142,89 @@ Has to be a valid UUID.
...
+NOT locate the actual TEA service endpoint at the
+".well-known" URI as per Section 1.1 of [RFC5785].
```
madpah, line 202:
```suggestion
`.well-known` URI as per Section 1.1 of [RFC5785].
```

discovery/readme.md
```diff
@@ -128,58 +142,89 @@ Has to be a valid UUID.
...
+### Overview: Finding the Index using DNS result
 
 Append the product part of the TEI to the URI found
 
 - TEI: `urn:tei:uuid:products.example.com:d4d9f54a-abcf-11ee-ac79-1a52914d44b1`
-- DNS record: `_tei._tcp.products.example.com`
-- URI in DNS: `://www.example.com/transparency/`
-- URL: `https://www.example.com/transparency/d4d9f54a-abcf-11ee-ac79-1a52914d44b1/`
+- DNS record: `products.example.com`
+- Server found in DNS HTTPS record: `teapot.prod.example.com`
```
madpah, line 210:
```suggestion
- Server found in DNS `HTTPS` record: `teapot.prod.example.com`
```

discovery/readme.md
```diff
@@ -128,58 +142,89 @@ Has to be a valid UUID.
...
+- URL: `https://products.example.com/.well-known/tea/d4d9f54a-abcf-11ee-ac79-1a52914d44b1/`
+- HTTP 302 redirect to "https://teapot02.consumer.example.com/tea/v2/product-index/d4d9f54a-abcf-11ee-ac79-1a52914d44b1'
+
+The server at "teapot.prod.example.com" needs a TLS certificate including "products.example.com"
```
madpah, line 214:
```suggestion
The server at `teapot.prod.example.com` needs a TLS certificate that valid for `products.example.com` 
```

discovery/readme.md
```diff
@@ -128,58 +142,89 @@ Has to be a valid UUID.
...
+in the subject alt name. "teapot02.consumer.example.com" needs a certificate with that
```
madpah, line 215:
```suggestion
 - e.g. with `products.example.com` in the subject alt name. `teapot02.consumer.example.com` needs a certificate with that
```

discovery/readme.md
```diff
@@ -128,58 +142,89 @@ Has to be a valid UUID.
...
+host name.
 
-If no DNS URI records are found the resolution defaults to A and AAAA records.
+If no HTTPS records are found the resolution defaults to A and AAAA records.
```
madpah, line 218:
```suggestion
If no `HTTPS` records are found the resolution defaults to `A` and `AAAA` records.
```

discovery/readme.md
oej: Removed all HTTPS support.

PR #75, opened 2024-11-18
Leaf: Clarify the prerelease-flag

### LINKED ISSUE #44: Can you promote a `pre-release` leaf?
A common workflow for `channel` like deployment flows is that you promote a build/version from say candidate to stable. In this case, you *may* want to maintain the UUID for the leaf.

oej:
Close #44

### DIFF
--- a/tea-leaf/tea-leaf.md
+++ b/tea-leaf/tea-leaf.md
@@ -3,7 +3,7 @@
 _Work in progress_
 
 The TEA LEAF is the object that indicates a product version. The API should be
-very agnostic to how a "version" is indicated - semver, vers, name, hash or anything else.
+very agnostic as to how a "version" is indicated - semver, vers, name, hash or anything else.
 
 ## Major and minor versions
 
--- a/tea-leaf/tea-leaf.md
+++ b/tea-leaf/tea-leaf.md
@@ -3,7 +3,7 @@
 _Work in progress_
 
 The TEA LEAF is the object that indicates a product version. The API should be
-very agnostic to how a "version" is indicated - semver, name, hash or anything else.
+very agnostic to how a "version" is indicated - semver, vers, name, hash or anything else.
 
 ## Major and minor versions
 
@@ -16,12 +16,23 @@ a timestamp for a release is required.
 
 ## The Leaf Object
 
-- __UUID__ unique for this object
+- __Uuid__ unique for this object
 - __Product name__: A text field
 - __Product version__: A text field, no required syntax
 - __Release date__: A unix timestamp
-- __Prerelease__: A flag indicating a pre-release (beta, rc)
-- __TCO_UUID__: A reference to the TEA Collection objet for this release
+- __Prerelease__: A boolean flag indicating a pre-release (beta, rc)
+- __Tco_uuid__: A reference to the TEA Collection objet for this release
+
+## Handling the Prerelease flag
+
+The Prerelease flag is used to indicate that this is not a final release. For a given Leaf with a UUID, the flag
+can be set to indicate a "test", "beta", "alfa" or similar non-deployed release. It can only be set when
+creating the LEAF. The TEA implementation may allow it to be unset (False) once. This is to support
+situations where a object is promoted as is after testing to production version. The flag can not
+be set after initial creation and publication of the leaf.
+
+If the final version is different from the pre-release (bugs fixed, code changed, different binary)
+a new leaf with a new UUID and version needs to be created.
 
 ## References
 
--- a/tea-leaf/tea-leaf.md
+++ b/tea-leaf/tea-leaf.md
@@ -20,7 +20,7 @@ a timestamp for a release is required.
 - __Product name__: A text field
 - __Product version__: A text field, no required syntax
 - __Release date__: A unix timestamp
-- __Prerelease__: A boolean flag indicating a pre-release (beta, rc)
+- __Pre-Release__: A boolean flag indicating a pre-release (beta, rc)
 - __Tco_uuid__: A reference to the TEA Collection objet for this release
 
 ## Handling the Prerelease flag
--- a/tea-leaf/tea-leaf.md
+++ b/tea-leaf/tea-leaf.md
@@ -1,7 +1,5 @@
 # The TEA Leaf Object (TLO)
 
-_Work in progress_
-
 The TEA LEAF is the object that indicates a product version. The API should be
 very agnostic as to how a "version" is indicated - semver, vers, name, hash or anything else.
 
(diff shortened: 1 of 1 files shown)


oej:
Ping @vpetersson

tea-leaf/tea-leaf.md
```diff
@@ -16,12 +16,23 @@ a timestamp for a release is required.
...
 - __Release date__: A unix timestamp
-- __Prerelease__: A flag indicating a pre-release (beta, rc)
-- __TCO_UUID__: A reference to the TEA Collection objet for this release
+- __Prerelease__: A boolean flag indicating a pre-release (beta, rc)
+- __Tco_uuid__: A reference to the TEA Collection objet for this release
+
+## Handling the Prerelease flag
+
+The Prerelease flag is used to indicate that this is not a final release. For a given Leaf with a UUID, the flag
+can be set to indicate a "test", "beta", "alfa" or similar non-deployed release. It can only be set when
```
ppkarwasz, line 29: Should we stress the fact that this applies to **non-deployed** releases, regardless of the naming.

For example [`org.apache.logging.log4j:log4j-core:3.0.0-beta3`](https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-core/3.0.0-beta3) does not count as pre-release, since we are not allowed by Maven Central to modify it.

oej: I would definitely set something named "-beta3" as a prerelease. If you modify something, that's always a new release with a new leaf. Maybe we need to be clearer in the text.

The thing with the flag change was that in some cases a pre-release is upgraded to final release after testing, without ANY changes. That's just another release behaviour.

madpah: @ppkarwasz makes a good point. We need to be clear on _our_ definition of what constitutes a pre-release in TEA.

madpah: The concept of "pre-release" will mean different things to different parties:

- Maven: As per example from @ppkarwasz - it _was_ published; it _cannot_ be modified; it is _not_ intended for use in Production systems
- COTS: E.g. Windows Server 3030 RC (fictitious) - it _was_ published; it _cannot_ be modified; it is _not_ intended for use in Production systems

For me both these examples would be considered as `pre-releases` by their Manufacturers.

oej: We have had long discussions about this before, but can of course restart if needed.

madpah: My only point really is that our definition is not clear in the docs we have. Not that it hasn't been discussed already 😉

vpetersson: Some thoughts on this:

* We need to be dynamic here -- in some workflows per-releases will not be published, but in other workflows they will (e.g. promoted). Thus, as long as we allow for `PATCH`ing the LEAF to toggle this boolean, we should be good and cover both workflows.
* NIT: `s/alfa/alpha/g` and `s/prerelease/pre-release/g`

tea-leaf/tea-leaf.md
```diff
@@ -3,7 +3,7 @@
 _Work in progress_
 
 The TEA LEAF is the object that indicates a product version. The API should be
-very agnostic to how a "version" is indicated - semver, name, hash or anything else.
+very agnostic to how a "version" is indicated - semver, vers, name, hash or anything else.
```
madpah, line 6:
```suggestion
very agnostic as to how a "version" is indicated - semver, vers, name, hash or anything else.
```

tea-leaf/tea-leaf.md
```diff
@@ -16,12 +16,23 @@ a timestamp for a release is required.
 
 ## The Leaf Object
 
-- __UUID__ unique for this object
+- __Uuid__ unique for this object
```
madpah, line 19: We've called this an `identifier` (which is of type UUID) in the OpenAPI Spec - we need to align in either direction.

tea-leaf/tea-leaf.md
```diff
@@ -16,12 +16,23 @@ a timestamp for a release is required.
...
 - __Product name__: A text field
 - __Product version__: A text field, no required syntax
 - __Release date__: A unix timestamp
-- __Prerelease__: A flag indicating a pre-release (beta, rc)
-- __TCO_UUID__: A reference to the TEA Collection objet for this release
+- __Prerelease__: A boolean flag indicating a pre-release (beta, rc)
```
madpah, line 23:
```suggestion
- __Pre-Release__: A boolean flag indicating a pre-release (beta, rc)
```

tea-leaf/tea-leaf.md
```diff
@@ -16,12 +16,23 @@ a timestamp for a release is required.
...
+- __Tco_uuid__: A reference to the TEA Collection objet for this release
+
+## Handling the Prerelease flag
```
madpah, line 26:
```suggestion
## Handling the Pre-Release flag
```

tea-leaf/tea-leaf.md
```diff
@@ -16,12 +16,23 @@ a timestamp for a release is required.
...
+
+The Prerelease flag is used to indicate that this is not a final release. For a given Leaf with a UUID, the flag
```
madpah, line 28:
```suggestion
The Pre-Release flag is used to indicate that this is not a final release. For a given Leaf with a UUID, the flag
```

tea-leaf/tea-leaf.md
madpah, line 29:
```suggestion
can be set to indicate a "test", "beta", "alpha" or similar non-deployed release. It can only be set when
```

madpah (changes requested):

tea-leaf/tea-leaf.md
madpah: Does everyone agree with @vpetersson's conclusion? Does that work for you @ppkarwasz?

I think we still need to add wording to clarify in Docs.

tea-leaf/tea-leaf.md
oej: Agree

tea-leaf/tea-leaf.md
oej: Seems like we have a conclusion. What text is needed to clarify more? I tried to be bit fuzzy in the text to not hardcode to a specific example. The flag is meant as a warning to not use this release. We can rename it to "not-recommended" or something along those lines to avoid the discussions about all variants of "beta", "pre-release", "test-release" etc. Just

tea-leaf/tea-leaf.md
oej: We can align in a later request

PR #77, opened 2024-11-19
Tweak OpenAPI specs and workflows

vpetersson:
* Completes OpenAPI specs with missing building blocks
* Synchronizes workflow diagram with API specs
* Minor documentation fixes

### DIFF
--- a/spec/openapi.json
+++ b/spec/openapi.json
@@ -176,7 +176,7 @@
                                     "type": "object",
                                     "properties": {
                                         "pagination": {
-                                            "$ref": "#/components/schemas/type-pagination-details"
+                                            "$ref": "#/components/schemas/type_pagination_details"
                                         },
                                         "items": {
                                             "type": "array",
--- a/spec/openapi.json
+++ b/spec/openapi.json
@@ -528,7 +528,24 @@
                                                                 "format": "int64"
                                                             },
                                                             "artifact_checksum": {
-                                                                "type": "string"
+                                                                "type": "string",
+                                                                "description": "The checksum value"
+                                                            },
+                                                            "artifact_checksum_type": {
+                                                                "type": "string",
+                                                                "description": "The algorithm used to generate the checksum",
+                                                                "enum": [
+                                                                    "SHA512",
+                                                                    "SHA384",
+                                                                    "SHA256",
+                                                                    "SHA224",
+                                                                    "SHA1",
+                                                                    "MD5",
+                                                                    "BLAKE2b-512",
+                                                                    "BLAKE2s-256",
+                                                                    "BLAKE3"
+                                                                ],
+                                                                "default": "SHA256"
                                                             },
                                                             "signature_url": {
                                                                 "type": "string",
@@ -540,7 +557,8 @@
                                                             "mime_type",
                                                             "artifact_url",
                                                             "artifact_size_in_bytes",
-                                                            "artifact_checksum"
+                                                            "artifact_checksum",
+                                                            "artifact_checksum_type"
                                                         ]
                                                     }
                                                 }
@@ -696,7 +714,24 @@
                                                                 "format": "int64"
                                                             },
                                                             "artifact_checksum": {
-                                                                "type": "string"
+                                                                "type": "string",
+                                                                "description": "The checksum value"
+                                                            },
+                                                            "artifact_checksum_type": {
+                                                                "type": "string",
+                                                                "description": "The algorithm used to generate the checksum",
+                                                                "enum": [
+                                                                    "SHA512",
+                                                                    "SHA384",
+                                                                    "SHA256",
+                                                                    "SHA224",
+                                                                    "SHA1",
+                                                                    "MD5",
+                                                                    "BLAKE2b-512",
+                                                                    "BLAKE2s-256",
+                                                                    "BLAKE3"
+                                                                ],
+                                                                "default": "SHA256"
                                                             },
                                                             "signature_url": {
                                                                 "type": "string",
@@ -708,7 +743,8 @@
                                                             "mime_type",
                                                             "artifact_url",
                                                             "artifact_size_in_bytes",
-                                                            "artifact_checksum"
+                                                            "artifact_checksum",
+                                                            "artifact_checksum_type"
                                                         ]
                                                     }
                                                 }
@@ -976,7 +1012,24 @@
                         "format": "int64"
                     },
                     "artifact_checksum": {
-                        "type": "string"
+                        "type": "string",
+                        "description": "The checksum value"
+                    },
+                    "artifact_checksum_type": {
+                        "type": "string",
+                        "description": "The algorithm used to generate the checksum",
+                        "enum": [
+                            "SHA512",
+                            "SHA384",
+                            "SHA256",
+                            "SHA224",
+                            "SHA1",
+                            "MD5",
+                            "BLAKE2b-512",
+                            "BLAKE2s-256",
+                            "BLAKE3"
+                        ],
+                        "default": "SHA256"
                     },
                     "signature_url": {
                         "type": "string",
@@ -989,7 +1042,8 @@
                     "mime_type",
                     "artifact_url",
                     "artifact_size_in_bytes",
-                    "artifact_checksum"
+                    "artifact_checksum",
+                    "artifact_checksum_type"
                 ]
             },
             "type_author": {
--- a/spec/openapi.json
+++ b/spec/openapi.json
@@ -25,7 +25,7 @@
     "paths": {
         "/product-index/{product-identifier}": {
             "get": {
-                "description": "Access the TEA Index for the supplied product identifier",
+                "description": "Returns the corresponding leafs for a given TEA Product Identifier",
                 "operationId": "getTeaIndex",
                 "parameters": [
                     {
@@ -40,7 +40,7 @@
                 ],
                 "responses": {
                     "200": {
-                        "description": "Requested TEA Index found and returned",
+                        "description": "Requested Product TEA Index found and returned",
                         "content": {
                             "application/json": {
                                 "schema": {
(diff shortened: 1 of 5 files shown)


vpetersson:
@oej / @stevespringett Please review. I realize this is a fairly sizeable PR with a lot of changes in here but my goal was to contribute a more complete API spec and bring it in line with the workflow documentation.

madpah:
@vpetersson - the OpenApi Spec JSON file is throwing errors for me (that don't make sense). Will take some time to confirm if they are indeed errors and if so what they are caused by!

vpetersson:
Interesting. Did you preview it with the swagger-ui in Docker as per the readme?

vpetersson:
Yeah I can't say I'm a big fan of swagger-ui myself, but it was the path of least resistance. There are more modern alternatives out there that we can adopt later on.

spec/openapi.json
```diff
@@ -23,13 +23,13 @@
         }
     ],
     "paths": {
-        "/product-index/{product-identifier}": {
+        "/product/{tea-product-identifier}": {
```
madpah, line 26: I'm happy with `/product`, but if I recall, @oej asked me to change it to `/product-index`.

Personally - I prefer this shortened version.

spec/openapi.json
```diff
@@ -40,43 +40,349 @@
...
+                        "$ref": "#/components/parameters/page-offset"
+                    },
+                    {
+                        "$ref": "#/components/parameters/page-size"
+                    },
+                    {
+                        "$ref": "#/components/parameters/purl"
+                    },
+                    {
+                        "$ref": "#/components/parameters/ean"
```
madpah, line 167: We need to be more inclusive that just EAN barcodes - UPC is the standard in US for example.

spec/openapi.json
```diff
@@ -40,43 +40,349 @@
...
+                        "basicAuth": []
+                    }
+                ]
+            }
+        },
+        "/product": {
+            "get": {
+                "description": "Returns a list of TEA Products",
+                "operationId": "listTeaProducts",
+                "parameters": [
```
madpah, line 156: Should we not have a more _vague_ API here? The intent is to return a list of TEA Products that match some identifier.

What if some supplies a value for each parameter (ignore pagination)? 

I'd collapse `purl`, `ean`, `sku` and `vendor-uuid` into a single `product-identifier` parameter.

spec/openapi.json
```diff
@@ -40,43 +40,349 @@
...
+                                }
+                            }
+                        }
+                    }
+                },
+                "tags": [
+                    "TEA Product"
+                ]
+            },
+            "post": {
```
madpah, line 207: Surely a `product-identifier` is required for the creation of a TEA Product?

spec/openapi.json
vpetersson: Yeah `/product-index` was inconsistent with the other endpoints. That's why I renamed it. But `/product` and `/tea` are indices technically speaking.

spec/openapi.json
vpetersson: Agreed. This needs a better name. I just pulled it [from here](https://github.com/CycloneDX/transparency-exchange-api/blob/main/doc/tea-requirements.md#repository-discovery). We also need something that includes newer style of GS1 barcodes.

spec/openapi.json
```diff
@@ -23,13 +23,13 @@
...
-                "operationId": "getTeaIndex",
+                "description": "Returns the corresponding leafs for a given TEA Product Identifier",
+                "operationId": "getTeaProduct",
                 "parameters": [
                     {
-                        "$ref": "#/components/parameters/product-identifier"
+                        "$ref": "#/components/parameters/tea-product-identifier"
                     },
                     {
                         "$ref": "#/components/parameters/page-offset"
```
madpah, line 35: Why is pagination required on this API when we are passing in a [unique] TEI?

spec/openapi.json
```diff
@@ -40,43 +40,349 @@
...
+                    {
+                        "bearerAuth": []
+                    },
+                    {
+                        "basicAuth": []
+                    }
+                ]
+            }
+        },
+        "/leaf": {
```
madpah, line 275: What is the value of the `/leaf` API endpoint? Who is calling this without specifying a `tea-leaf-identifier` ?

spec/openapi.json
```diff
@@ -175,28 +946,40 @@
...
+                        "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?(?:Z|[+-]\\d{2}:?\\d{2})$",
+                        "example": "2024-03-20T15:30:00Z"
+                    },
+                    "collection-references": {
+                        "type": "array",
+                        "items": {
+                            "$ref": "#/components/schemas/type-uuid"
+                        },
+                        "description": "Array of UUIDs referencing TEA Collection entries",
+                        "default": []
```
madpah, line 968: An `example` would be helpful in this case

spec/openapi.json
```diff
@@ -207,7 +990,9 @@
                     },
                     "release-date": {
                         "type": "string",
-                        "format": "date-time"
+                        "format": "date-time",
+                        "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?(?:Z|[+-]\\d{2}:?\\d{2})$",
```
madpah, line 994: Is this `pattern` required? `date-time` specifically implies `2018-11-13T20:20:39+00:00` format.

See https://json-schema.org/understanding-json-schema/reference/string#dates-and-times

spec/openapi.json
```diff
@@ -107,6 +413,285 @@
...
+                    {
+                        "bearerAuth": []
+                    },
+                    {
+                        "basicAuth": []
+                    }
+                ]
+            }
+        },
+        "/collection": {
```
madpah, line 516: Unclear the value of `/collection` API endpoint - why would this be called without `tea-collection-identifier` ever?

spec/openapi.json
```diff
@@ -107,6 +413,285 @@
...
+                                        "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?(?:Z|[+-]\\d{2}:?\\d{2})$",
+                                        "example": "2024-03-20T15:30:00Z"
+                                    },
+                                    "author": {
+                                        "$ref": "#/components/schemas/type-author"
+                                    },
+                                    "reason": {
+                                        "$ref": "#/components/schemas/type-collection-reason"
+                                    },
+                                    "artifacts": {
```
madpah, line 590: This object would benefit from an `example`.

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                },
+                                                "objects": {
+                                                    "type": "array",
+                                                    "items": {
+                                                        "type": "object",
+                                                        "properties": {
+                                                            "description": {
+                                                                "type": "string"
+                                                            },
+                                                            "mime_type": {
```
madpah, line 780: Snake_Case instead of kebab-case

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                                "type": "string"
+                                                            },
+                                                            "artifact_url": {
```
madpah, line 783: Snake_Case instead of kebab-case

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                                "type": "string",
+                                                                "format": "uri"
+                                                            },
+                                                            "artifact_size_in_bytes": {
```
madpah, line 787: Snake_Case instead of kebab-case

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                                "type": "integer",
+                                                                "format": "int64"
+                                                            },
+                                                            "artifact_checksum": {
```
madpah, line 791: Checksum - what type? Are we supporting a single checksum algorithm? Or multiple?

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                                "type": "string"
+                                                            },
+                                                            "signature_url": {
```
madpah, line 794: Snake_Case instead of kebab-case

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                                "type": "string",
+                                                                "format": "uri"
+                                                            }
+                                                        },
+                                                        "required": [
+                                                            "description",
+                                                            "mime_type",
```
madpah, line 801: Snake_Case instead of kebab-case

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                            "artifact_url",
```
madpah, line 802: Snake_Case instead of kebab-case

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                            "artifact_size_in_bytes",
```
madpah, line 803: Snake_Case instead of kebab-case

spec/openapi.json
```diff
@@ -136,35 +721,221 @@
...
+                                                            "artifact_checksum"
```
madpah, line 804: Snake_Case instead of kebab-case

spec/openapi.json
```diff
@@ -107,6 +413,285 @@
...
+                                                },
+                                                "type": {
+                                                    "$ref": "#/components/schemas/type-tea-collection-artifact-type"
+                                                },
+                                                "author": {
+                                                    "$ref": "#/components/schemas/type-author"
+                                                },
+                                                "objects": {
+                                                    "type": "array",
+                                                    "items": {
```
madpah, line 606: Refer to `tea-collection-artifact-object`

spec/openapi.json
vpetersson: Isn't this a UUID that is created by the server, rather than something that is user provided? Products, Leafs and Collections will all receive a unique UUID from the server when created.

spec/openapi.json
vpetersson: That's a good point. My initial thinking was that it would display *all* leafs, but you're right, that is probably not super useful.

spec/openapi.json
vpetersson: Yeah perhaps loosen this is fine.

spec/openapi.json
madpah: Will open GH issue to track this.

spec/openapi.json
vpetersson: Yeah same as for `/leaf`: listing. Will drop both.

spec/openapi.json
madpah: https://github.com/CycloneDX/transparency-exchange-api/issues/78

spec/openapi.json
madpah: I mean - what about the Product EAN, Part Code or such that is referenced in other parts of the API?

vpetersson: I consider that metadata. A product could have both a SKU, EAN Barcode and other values. The UUID is the identifier in the system.

spec/openapi.json
vpetersson: Added this. But will probably need to be overhauled.

vpetersson: Yep, this was in the previous version. I thought the same thing. Added initial support for this.

spec/openapi.json
madpah: Raised https://github.com/CycloneDX/transparency-exchange-api/issues/79

spec/openapi.json
madpah: Being tracked in https://github.com/CycloneDX/transparency-exchange-api/issues/80

spec/openapi.json
madpah: Understood @vpetersson - but that leaves us, with the API Spec as is, not being able to _create_ said metadata or add any of these `product-identifiers`.

Will open GH Issue to track

madpah: Created https://github.com/CycloneDX/transparency-exchange-api/issues/81

madpah (approved):

madpah:
All comments addressed in the PR or GH Issues opened to track resolution beyond this PR.

- #78 
- #79 
- #80 
- #81

PR #82, opened 2024-11-20
fix: `examples` incorrectly nested for `POST /leaf`

madpah:
Root cause of https://github.com/madpah/transparency-exchange-api/actions/runs/11933709569/job/33261312468


madpah:
Reviewed by @vpetersson (not yet maintainer).

PR #83, opened 2024-11-20
Adds the ability to patch `/product/tea_product_identifier`

### LINKED ISSUE #81: SPEC: `/product` endpoints - incomplete
`GET /product` returns a paginated list of `#/components/schemas/tea_product`, which includes fields such as `purl` and `barcode` currently.

However, there is no mechanism provided in the API Spec as it stands to set/provide fields for a Product beyond it's Name in `POST /product`

From: https://github.com/CycloneDX/transparency-exchange-api/pull/77#discussion_r1850139350

FYI @oej  @vpetersson

vpetersson:
I forgot to add this in the previous PR.

Closes #81.

### DIFF
--- a/spec/openapi.json
+++ b/spec/openapi.json
@@ -235,6 +235,22 @@
                                 "properties": {
                                     "product_name": {
                                         "type": "string"
+                                    },
+                                    "barcode": {
+                                        "type": "string",
+                                        "description": "Barcode"
+                                    },
+                                    "sku": {
+                                        "type": "string",
+                                        "description": "Product SKU"
+                                    },
+                                    "vendor_uuid": {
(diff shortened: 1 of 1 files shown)


spec/openapi.json
```diff
@@ -70,11 +70,36 @@
                                 "properties": {
                                     "product_name": {
                                         "type": "string"
+                                    },
```
madpah, line 73: Did we drop the `required` for `product_name` @vpetersson ?

vpetersson: Hmm, i think `product_name` should be required as it's the human readable name for the product.

madpah (approved):

PR #84, opened 2024-11-20
DRY up the file with reusable blocks.

vpetersson:


madpah:
Am reviewing @vpetersson - I note we have a root level `security:` block, but also `security` sections at operation level - I don't think both are required...

vpetersson:
Thanks for catching! Yes, reckless refactoring from me. Thanks for finding that. Fixing.

madpah:
@vpetersson Line 66 (`PATCH /product/{tea_product_identifier} requestBody` could use `#/schemas/compoennts/base_product_fields`

madpah:
An example of another area we can consider schema consolidation:

`GET /product parameters` could make use of `allOf` if we create a Schema Object with the Pagination Parameters - as these are re-used in a few places.

Thoughts @vpetersson ?

oej:
Seems like we're getting close here.

oej:
Moved @madpah comment to a separate issue. Will go ahead and merge this and we'll open new issues and PRs to move forward.

Issue #97, opened 2025-02-12
Document authentication and authorisation


oej:
<img width="1198" alt="Image" src="https://github.com/user-attachments/assets/7d52fcfc-b5d3-4f15-897c-e08064082702" />

oej:
Done

oej:
We have a document. Comments, changes will be another issue

PR #98, opened 2025-02-12
Adding first draft of authentication/authorization documentation

oej:
Closes issue #97

### DIFF
--- a/auth/readme.md
+++ b/auth/readme.md
@@ -1,0 +1,42 @@
+# Transparency Exchange API - Authentication and authorization
+
+
+**NOTE**: _This is a proposal for the WG_
+
+This document covers authentication and authorization on the consumer side
+of a TEA service - the discovery and download of software transparency artefacts.
+
+## Requirements
+
+__Authorization__: A user of a TEA service may get access to all objects (leaf, collections) and
+artefacts or just a subset, depending on the publisher of the data. Authorization is connected
+to __authentication__. 
+
+The level of authorization is up to the implementer of the TEA implementation and the publisher.
+
+In order to get interoperability between clients and servers implementing the protocol, the
+specification focuses on the authentication. After successful authentication, the authorization
+may be implemented in multiple ways - on various levels of the API - depending on what information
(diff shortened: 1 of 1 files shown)


oej:
Add http responses for various cases:
- Expired token
- Auth revoked
- No access to product/artefact

auth/readme.md
```diff
@@ -0,0 +1,42 @@
...
+__Authorization__: A user of a TEA service may get access to all objects (leaf, collections) and
+artefacts or just a subset, depending on the publisher of the data. Authorization is connected
+to __authentication__. 
+
+The level of authorization is up to the implementer of the TEA implementation and the publisher.
+
+In order to get interoperability between clients and servers implementing the protocol, the
+specification focuses on the authentication. After successful authentication, the authorization
+may be implemented in multiple ways - on various levels of the API - depending on what information
+the user can access.
```
ppkarwasz, line 20: Should we specify that authorization should not require additional work on the part of an authenticated client?

What I mean is that clients should have a **single** bearer token to authenticate to the server, instead of separate tokens for each resource they request.

auth/readme.md
```diff
@@ -0,0 +1,42 @@
...
+versions openly, but restrict access to artefacts to those that match the customers installation.
+Another implementation can implement a filter that does not show products and versions ("leafs") that
+the customer has not aquired.
+
+For most Open Source projects, implementing authentication - setting up accounts and managing
+authorization - does not make much sense, since the information is usually in the open any way.
+
+## HTTP bearer token auth
+
+The API will support HTTP bearer token in the __Authorization:__ http header.
```
vpetersson, line 32: I would personally keep this out of scope. In some machine-to-machine scenarios, using Mutual TLS (mTLS) may make more sense.  Thus I would suggest leaving out this block in favor of just saying that it is up to the implementor.

oej: We are trying to limit the number of auth schemes to create interoperability. I will add text that either the token or TLS client certs may be used, it's good idea.

PR #100, opened 2025-02-17
Fix Open API spec

RingoDev:
Hey @oej this is a fix up of some of the broken pieces of @vpetersson's WIP PR.

Discussed here: https://cyclonedx.slack.com/archives/C04LR6R9T8E/p1739278644285349

### DIFF
--- a/spec/openapi.json
+++ b/spec/openapi.json
@@ -43,11 +43,9 @@
                             }
                         }
                     },
-
                     "404": {
                         "$ref": "#/components/responses/404-object-by-id-not-found"
                     }
-
                 },
                 "tags": [
                     "TEA Product"
@@ -121,21 +119,15 @@
                             "application/json": {}
                         }
                     },
-                    "401": { "$ref": "#/components/responses/401-unauthorized" },
+                    "401": {
+                        "$ref": "#/components/responses/401-unauthorized"
+                    },
                     "404": {
                         "$ref": "#/components/responses/404-object-by-id-not-found"
                     }
                 },
                 "tags": [
                     "TEA Product"
-                ],
-                "security": [
-                    {
-                        "bearerAuth": []
-                    },
-                    {
-                        "basicAuth": []
-                    }
                 ]
             },
             "delete": {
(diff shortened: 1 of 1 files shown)


spec/openapi.json
```diff
@@ -146,7 +146,19 @@
...
+                    "204": {
+                        "description": "Resource successfully deleted"
+                    }
+                },
+                "security": [
+                    {
+                        "bearerAuth": []
+                    },
+                    {
+                        "basicAuth": []
```
oej, line 159: Why do we have basic auth here?

RingoDev: Good question, that is already in there for a few months: https://github.com/CycloneDX/transparency-exchange-api/blame/main/spec/openapi.json#L369

RingoDev: I guess it should be removed?

oej: Ok, let's just remove basicAuth

RingoDev: According to this the security requirements do not need to be specified at all per operation: https://learn.openapis.org/specification/security.html

We already have global definitions here: https://github.com/CycloneDX/transparency-exchange-api/pull/100/files#diff-e86fbd38497786da8d475549c2d7c81274c747efed425ef817e8df8aaa0a5004R1473-R1480

RingoDev:
I removed the _Security Requirement_ we had on each operation as this is set globally and the same for all endpoints (at the moment)
I also got rid of the basic-auth security scheme definition and global requirement.

RingoDev:
Last point to note: To fix the problems with the spec I had to make an assumption about how the "Product Index" (GET /product) should be paginated.
It is now wrapped in the way that the endpoint returns an object with pagination attributes at top level as well as a `content` attribute that contains the list of products.

```
"paginated_product_list": {
    "type": "object",
    "allOf": [
        {
            "$ref": "#/components/schemas/type_pagination_details"
        },
        {
            "type": "object",
            "properties": {
                "content": {
                    "type": "array",
                    "items": {
                        "$ref": "#/components/schemas/tea_product"
                    }
                }
            }
        }
    ]
}
```

vpetersson (approved):
Nice cleanup.

ppkarwasz:
@oej,

Since #119 does not have validation errors, should we close this PR and merge that one?

PR #104, opened 2025-03-10
Update openapi spec

oej:
- remove publication API from core spec
- start work with unification of specifications


ppkarwasz (approved):
Let's merge it and work on the splitted versions

ppkarwasz (approved):

vpetersson (approved):

PR #105, opened 2025-03-13
Adds Markdown Linter

vpetersson:
* Adds Markdown linter
  * I used line lenght of 80, which appears to be what @oej has been using 
* Reformatted all Markdown files (which found a few errors too)


oej:
Brilliant! I like having a linter and since we have markdown now, I think it's a good idea. I'll take a more detailed look later this weekend.

madpah:
@vpetersson - any reason you have chosen not to use an existing GitHub Action for this?
- https://github.com/DavidAnson/markdownlint-cli2-action
- https://github.com/super-linter/super-linter

PR #113, opened 2025-04-16
Cherrypick out linter

vpetersson:
Cherry pick out Markdown linter without actually running it.


madpah:
@vpetersson - any reason you have chosen not to use an existing GitHub Action for this?

- https://github.com/DavidAnson/markdownlint-cli2-action
- https://github.com/super-linter/super-linter

vpetersson:
@madpah I've found Super Linter to be super slow (and bloated) in the past. I just copied this from other pipelines where I've added this setup. We could definitely add an upstream Action, but it was just a few lines so i figured it was easier to implement.

madpah:
No strong opinion - just having the chat 😉

vpetersson:
All good, mate! Yeah I usually try to use (well maintained) Action modules too, but haven't found a good one for this.

Also, note that the tests are failing because I didn't run it. @oej asked me to *not* apply the linting in this PR.

oej:
Hmm. This fails because of the workflow with the linter... Isn't that weird or should it be that way? Anything we have to do before merging?

PR #114, opened 2025-04-17
Modifies domain objects after workshop

ppkarwasz:
This PR modifies the schema of the domain objects according to the structure discussed in the Koala Workshop

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -48,8 +48,8 @@ paths:
         match.
       operationId: getTeaProductByIdentifier
       parameters:
-        - $ref: "#/components/parameters/page_offset"
-        - $ref: "#/components/parameters/page_size"
+        - $ref: "#/components/parameters/page-offset"
+        - $ref: "#/components/parameters/page-size"
         - $ref: "#/components/parameters/tei_urn"
       responses:
         '200':
@@ -60,7 +60,7 @@ paths:
           $ref: "#/components/responses/404-object-by-id-not-found"
       tags:
         - TEA Product
-  /tea-component/{uuid}:
+  /component/{uuid}:
     get:
       description: Get the TEA Leaf that describes the Version of a Product
       operationId: getTeaComponentById
@@ -85,7 +85,7 @@ paths:
           $ref: "#/components/responses/404-object-by-id-not-found"
       tags:
         - TEA Leaf
-  /component/{uuid}/collections:
+  /component/{uuid}/versions:
     get:
       description: Get the TEA Leaf that describes the Version of a Product
       operationId: getCollectionsByComponentId
@@ -105,7 +105,7 @@ paths:
               schema:
                 type: array
                 items:
-                  "$ref": "#/components/schemas/tea_collections_list_view"
+                  "$ref": "#/components/schemas/component-version"
         '400':
           $ref: "#/components/responses/400-invalid-request"
         '404':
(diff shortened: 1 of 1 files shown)


spec/openapi.yaml
```diff
@@ -82,274 +82,205 @@ paths:
...
+        - name
+        - identifiers
+        - components
+
+    #
+    # TEA Component and related objects
+    #
+    component:
       type: object
+      description: A TEA component
```
ppkarwasz, line 162: One of the possibilities discussed in the Workshop was to:

- inline the direct properties of this component (except `versions`) into `product.components`.
- have an API endpoint that returns a paginated list of `component_version` objects.

spec/openapi.yaml
```diff
@@ -82,274 +82,205 @@ paths:
...
         - name
-        - type
-        - author
-        - objects
-    tea_collection_artifact_object:
+        - identifiers
+        - versions
+    component_version:
       type: object
+      description: A specific version of a TEA leaf
```
ppkarwasz, line 186: A `component_version` describes a specific binary file that is distributed to consumers. Should we also have a checksum or that file?

Should we have an API endpoint to find the correct `component_version` based on its checksum?

ppkarwasz:
This PR is based on #117.

ppkarwasz:
Since there are lot of changes in this PR, here is a breakdown on how this integrates with the API Flow:

1. If a client knows the UUID of a TEA product, it can call `/product/{uuid}` to obtain such an object:
   https://github.com/CycloneDX/transparency-exchange-api/blob/a805f3dcda362e6d988399754813908d0bc24e43/spec/openapi.yaml#L174-L192

2. If the client does **not** know the UUID of a TEA product, it can call `/products` with some query parameters and get a paginated list of the same objects as above.
3. Once the client has a list of UUIDs of TEA components, it probably should call `/component/{uuid}` to identify what those UUIDs correspond to. We probably can optimize the workflow by including the object below in the answer returned by `/product/{uuid}`: 
   https://github.com/CycloneDX/transparency-exchange-api/blob/462b195c353378ab0372650ec8ed2e91dca15204/spec/openapi.yaml#L202-L215 

4. Once the client determines which component it is interested in it can call `/component/{uuid}/versions` with some query parameters (currently _none_) to list of the available versions of the component:
   https://github.com/CycloneDX/transparency-exchange-api/blob/a805f3dcda362e6d988399754813908d0bc24e43/spec/openapi.yaml#L226-L248

5. The last call is a call to `/collection/{uuid}`, which returns the **latest** version of the collection with a list of artifacts. Each artifact contains a list of artifact formats:
   https://github.com/CycloneDX/transparency-exchange-api/blob/a805f3dcda362e6d988399754813908d0bc24e43/spec/openapi.yaml#L259-L275

oej (changes requested):

spec/openapi.yaml
```diff
@@ -76,16 +78,16 @@ paths:
...
+          $ref: "#/components/responses/404-object-by-id-not-found"
       tags:
-        - TEA Leaf
-  /component/{uuid}/collections:
+        - TEA Component
+  /component/{uuid}/versions:
     get:
-      description: Get the TEA Leaf that describes the Version of a Product
+      description: Get all TEA Component versions that match the given criteria
       operationId: getCollectionsByComponentId
```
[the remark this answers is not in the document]
oej: Being fixed separately

PR #127, opened 2025-05-02
feat: add release into spec, further clean up

taleodor:


spec/openapi.yaml
```diff
@@ -218,36 +301,25 @@ components:
...
+        uuid:
+          "$ref": "#/components/schemas/type_uuid"
         version:
           type: string
-          description: Version number of TEA component
-        identifiers:
-          type: array
-          description: List of identifiers for this component version
-          items:
-            "$ref": "#/components/schemas/identifier"
```
ppkarwasz, line 234: The Package URL of a TEA Component and a TEA Component Release differ by **at least** the `version` part, so a TEA Component Release **should** also have a set of identifiers.

A TEA Component could have a PURL like `pkg:maven/groupId/artifactId`, while a TEA Component Release might have `pkg:maven/groupId/artifactId@1.0.0` or `pkg:maven/groupId/artifactId@1.0.0?classifier=shaded`.

spec/openapi.yaml
```diff
@@ -218,36 +301,25 @@ components:
...
-          description: Release date
-        pre_release:
-          type: boolean
-          description: Marks if the version is a pre-release version
-        classifier:
-          type: object
-          description: Optional classifier to distinguish between artifacts
+          type: string
+          format: date-time
+          example: '2024-03-20T15:30:00Z'
```
ppkarwasz, line 317:
```suggestion
          "$ref": "#/components/schemas/date-time"
```

The idea is to use a common timestamp format that complies both to [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) and ISO 8601.

spec/openapi.yaml
```diff
@@ -510,6 +493,102 @@ components:
...
+      #     of a written signature or a stylized script of the person's name.
+      #   digital-signature: A signature that leverages cryptography, typically public/private
+      #     key pairs, which provides strong authenticity verification.
+      #   rfc-9116: Document that complies with RFC-9116 (A File Format to Aid in Security
+      #     Vulnerability Disclosure)
+      #   other: Use this if no other types accurately describe the purpose of the external
+      #     reference.
+    type_uuid:
+      type: string
+      format: uuid
```
ppkarwasz, line 591:
```suggestion
```

There is already a `uuid` type above.

spec/openapi.yaml
```diff
@@ -218,36 +301,25 @@ components:
...
-            "$ref": "#/components/schemas/identifier"
         release_date:
-          "$ref": "#/components/schemas/date-time"
-          description: Release date
-        pre_release:
-          type: boolean
-          description: Marks if the version is a pre-release version
-        classifier:
-          type: object
-          description: Optional classifier to distinguish between artifacts
```
ppkarwasz, line 243: Why do we remove these?

spec/openapi.yaml
```diff
@@ -99,22 +97,74 @@ paths:
...
+        '400':
+          $ref: "#/components/responses/400-invalid-request"
+        '404':
+          $ref: "#/components/responses/404-object-by-id-not-found"
+      tags:
+        - TEA Component
+  /release/{uuid}/collection:
+    get:
+      description: Get the latest TEA Collection belonging to the TEA Release
+      operationId: getLatestCollectionByReleaseId
```
ppkarwasz, line 116: This endpoint is interesting, but I would rather call it `/collection/{uuid}/latest` or `/collection/{uuid}/current`. There is no risk of collision with `/collection/{uuid}/{version}`, since `{version}` is an integer.

spec/openapi.yaml
```diff
@@ -367,99 +443,6 @@ components:
         - digital-signature
         - rfc-9116
         - other
-      meta:enum:
-        vcs: Version Control System
-        issue-tracker: Issue or defect tracking system, or an Application Lifecycle
-          Management (ALM) system
-        website: Website
```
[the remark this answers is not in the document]
taleodor: I believe I commented this out, not removed completely, the tools I'm using are complaining otherwise.

spec/openapi.yaml
```diff
@@ -510,6 +493,102 @@ components:
         - BLAKE2b-512
         - BLAKE3
 
+      # meta:enum:
+      #   vcs: Version Control System
+      #   issue-tracker: Issue or defect tracking system, or an Application Lifecycle
+      #     Management (ALM) system
+      #   website: Website
```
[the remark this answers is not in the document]
taleodor: Then we should remove, and move to docs, agree.

spec/openapi.yaml
```diff
@@ -99,22 +97,74 @@ paths:
...
+        '400':
+          $ref: "#/components/responses/400-invalid-request"
+        '404':
+          $ref: "#/components/responses/404-object-by-id-not-found"
+      tags:
+        - TEA Release
+  /release/{uuid}/collections:
+    get:
+      description: Get the TEA Collections belonging to the TEA Release
+      operationId: getCollectionsByReleaseId
```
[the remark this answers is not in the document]
taleodor: This is for different collection versions, while /collection/{uuid} endpoint is for the latest version of the collection.

spec/openapi.yaml
taleodor: We can have both in theory; it is also possible that a collection belongs to several releases.

spec/openapi.yaml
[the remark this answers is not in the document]
ppkarwasz: Let's move it to the Markdown documentation.

spec/openapi.yaml
ppkarwasz: In most cases I don't think a TEA Collection can be shared between TEA Component Releases.

As soon as a TEA Collection contains an SBOM, the SBOM will be for a specific PURL. Different PURLs in my understanding correspond to different TEA Component Releases. Most of the TEA Artifacts can be shared between TEA Collections, so this is not a problem.

taleodor: There is a difference though between /release/{uuid} and /collection/{uuid} as first assumes that you know release's uuid and second assumes that you know collection's uuid - and those are not the same thing in the current paradigm. This is from pure API prospective. Generally, we need @oej in this conversation as personally I am not even in favour of having collection a first class object.

spec/openapi.yaml
taleodor: Those got removed by accident as we were introducing release object in parallel. However, personally, I would prefer a larger conversation on lifecycle, rather than introducing pre_release as a boolean flag. Concept of classifier here is unclear to me at all. I think we should have a call on this.

Personally, I would prefer removing this for Beta 1 with a note to introduce proper lifecycle management later.

oej: The pre-release flag has been discussed in detail in many meetings and needs to stay. The reason is that some releases get promoted from beta/pre-release to production without any change. If you release a separate pre-release and then change the code, that is two different releases in tea. 

We can open this discussion again after beta, but not before, in my view.

spec/openapi.yaml
oej: It is possible that artefacts belong to multiple collections, but not that a collection belongs to multiple releases. A collection is what a user keep checking - if I use version 3.4.3 of the OlleBlaster I need to check if there are any changes, a new version of the collection.

ppkarwasz: > I am not even in favour of having collection a first class object.

In general I agree with this: TEA Collection is not a first class object that should have its own UUID. To me it seems more like a property of the TEA Component Release project.

Therefore `/release/{uuid}/collection` and `/release/{uuid}/collection/{version}` seem more natural than `/collection/{uuid}` and `/collection/{uuid}/{version}`.

taleodor: I like this suggestion. Essentially, we could then treat collections as lamport timestamps (for my use case). Still a complication but much more manageable. I believe we should still iron this out on a meeting.

spec/openapi.yaml
```diff
@@ -1,5 +1,3 @@
----
-jsonSchemaDialect: https://spec.openapis.org/oas/3.1/dialect/base
```
oej, line 2: Why remove this? It was recently added if I remember correctly. Maybe @ppkarwasz has input

taleodor: This is not a valid key for YAML schema.

taleodor: Hm, I got this flagged by a linter, guess it was not updated or something.

ppkarwasz: Maybe we need to add a `$schema` property with value:

```yaml
$schema: https://spec.openapis.org/oas/3.1/schema-base/2025-02-13
```

spec/openapi.yaml
taleodor: Moving classifier conversation to issue 131.

spec/openapi.yaml
taleodor: Resolving, #132 for enhancing the return view of collections.

ppkarwasz (approved):

PR #137, opened 2025-05-07
Clarify TEA Collection fields

ppkarwasz:
This PR:

- Updates the fields in the `tea-collection.md` file based on the OpenAPI spec of the TEA Collection object.
- Synchronizes the descriptions between OpenAPI and Markdown.
- Adds an example of TEA Collection (real example of `log4j-core`).

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -366,17 +366,25 @@ components:
       description: A collection of security-related documents
       properties:
         uuid:
+          description: |
+            UUID of the TEA Collection object.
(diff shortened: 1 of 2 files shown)


spec/openapi.yaml
```diff
@@ -483,19 +531,19 @@ components:
       type: object
       description: A security-related document in a specific format
       properties:
-        mimeType:
+        mime_type:
```
ppkarwasz, line 534: Renamed to follow the snake_case convention in other properties.

PR #136, opened 2025-05-07
Clarify TEA Component Release fields

ppkarwasz:
- Fixes the fields in the TEA Component Release object, based on those listed in the OpenAPI spec.
- Synchronizes the descriptions used in `tea-collection.md` and the OpenAPI spec.
- Adds four examples of releases of the same TEA Component (Apache Tomcat) that differ in packaging, version numbers and presence of the pre-release flag.

PR #138, opened 2025-05-07
feat: use `camelCase` for property and parameter names

ppkarwasz:
This PR tries to uniformize the naming strategy for JSON properties and parameter names. Since CycloneDX mostly uses `camelCase` instead of `snake_case` or `kebab-case`, I settled to convert everything to camel case.

**Note**: The names that are used internally (e.g., JSON schemas and parameter references) were not modified.


ppkarwasz:
#136 might be a blocker, since it contains `release_date`.

PR #140, opened 2025-05-08
feat: find products by identifier

ppkarwasz:
This change replaces the `tei_urn` query parameter with a pair of `idType/idValue` parameters to filter products using all the available identifiers.

The ideal solution in this case would be to use a single parameter with schema `#/components/schemas/identifier` together with [RFC 6570-style serialization](https://swagger.io/specification/#style-examples):

```yaml
identifier-param:
  name: identifierParam
  description: If present, only the objects with the given identifier will be returned.
  in: query
  schema:
    $ref: "#/components/schemas/identifier"
  style: form
  explode: true
```

This is currently not a viable option, since limitations of the `openapi-generator` like OpenAPITools/openapi-generator#4808 would cause bugs in the generated clients.

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -1,3 +1,4 @@
+# $schema: https://spec.openapis.org/oas/3.1/schema-base/2025-02-13
 openapi: 3.1.1
 jsonSchemaDialect: https://spec.openapis.org/oas/3.1/dialect/base
 info:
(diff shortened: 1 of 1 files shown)


spec/openapi.yaml
```diff
@@ -680,15 +682,45 @@ components:
...
+    #     $ref: "#/components/schemas/identifier"
+    #   style: form
+    #   explode: true
+    #
+    # In the meantime we explode the object manually:
+    id-type:
+      # To allow RFC 6570 in the future without breaking changes to the HTTP API,
+      # the name of this parameter should be identical to the equivalent property in /components/schemas/identifier
+      name: idType
+      description: Type of identifier specified in the `idValue` parameter
```
oej, line 711: We need an enum to limit the number of types

ppkarwasz: This parameter already uses the `identifier-type` enum. This is defined two lines below:

https://github.com/CycloneDX/transparency-exchange-api/blob/6607f93f03036fd24549266870d53cb3d409ee84/spec/openapi.yaml#L713-L714

PR #141, opened 2025-05-08
fix: OpenAPI spec fixes as discussed on community meeting on 2025-05-07

taleodor:


spec/openapi.yaml
```diff
@@ -405,8 +405,8 @@ components:
           description: |
             TEA Collection version, incremented each time its content changes.
             Versions start with 1.
-        releaseDate:
-          description: TEA Collection version release date.
+        date:
+          description: The date when the TEA Collection version was created.
```
ppkarwasz, line 409: The word `date` doesn't suggest to me what kind of data it is. Maybe we could use `created`?

[CycloneDX uses `timestamp`](https://cyclonedx.org/docs/1.6/json/#metadata_timestamp), so `timestamp` should be OK too.

ppkarwasz (changes requested):

PR #142, opened 2025-05-08
Remove required prop for tei urn param and fix integers

taleodor:


taleodor:
PR bugged, reopening on anew one.

PR #146, opened 2025-05-09
feat: trim down TEA Artifact types

ppkarwasz:
Many of the [CycloneDX external reference](https://cyclonedx.org/docs/1.6/json/#externalReferences_items_type) types, such as `website` are not very interesting from a security perspective. Since it is easier to add new types than to remove them, this PR trims down the current list to some categories that:

- Should be interesting to access on a TEA server.
- Have machine-readable formats (with a few exceptions).

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -498,49 +498,17 @@ components:
       type: string
       description: Specifies the type of external reference.
       enum:
-        - vcs
-        - issue-tracker
-        - website
-        - advisories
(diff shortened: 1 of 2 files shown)


ppkarwasz:
This list contains the kinds of documents that _might_ be present on a TEA server. If a category is unlikely to be interesting, let us remove it. We can extend this enum at any time.

taleodor:
Looks great!

PR #177, opened 2025-05-28
feat: add TEA Release formats

ppkarwasz:
Similarly to TEA Artifacts, each TEA Component Release can include multiple deliverables representing the same software in different file formats or for different architectures.

This PR introduces:

- A `formats` property to the TEA Component Release, which includes:
  - a description,
  - a set of identifiers,
  - for software components: a set of checksums for the binary distribution file, a download URL, and a URL for an external signature.
- A `componentFormats` property to TEA Artifact, allowing restriction of which component formats a document applies to.

These changes enable more precise mapping between security artifacts and the distribution formats of components.

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -360,26 +360,62 @@ components:
           description: List of identifiers for the component
           items:
             "$ref": "#/components/schemas/identifier"
+        formats:
+          type: array
+          description: List of different formats of this component release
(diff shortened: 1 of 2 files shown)


taleodor (approved):
I would approve this in the current state. I believe we also need enums for specific architectures, CPU types, etc, but that could be added later.

PR #180, opened 2025-05-30
feat: This is a simplification of the Consumer API Spec

madpah:
This is draft PR that follows the Slack discussions [here](https://cyclonedx.slack.com/archives/C04LR6R9T8E/p1748619275186669) and [here](https://cyclonedx.slack.com/archives/C04LR6R9T8E/p1748619372215589) - supported by the attached Zoom Whiteboard  (also [here](https://zoom.us/wb/doc/MAXZhgSwQH2Yuyvqjd3p8Q/p/141222146461318)) where @nscuro and myself considered, from the Consumer perspective a simplification whilst also consider (as discussed during OWASP Hackathon) the necessary support for Product Versioning.

### Revised API Flow (Main Flow)

```mermaid
sequenceDiagram
    actor Client
    Client->>TEA Service: POST /product/search with TEI
    TEA Service-->>Client: TEA Product returned
    Note over TEA Service: This includes a list of TEA Collection identifiers

    loop For every TEA Collection identifier
    Client->>TEA Service: GET /collection/{id}/latest
    TEA Service-->>Client: TEA Collection returned
    Note over TEA Service: This includes all TEA Artifacts in the TEA Collection
    end
```

### References / Workings

Revised simplified data model suggestion (data model, not API Spec):
Consumer API Spec Review - OWASP Hackathon_Page 3- 1748631550990

Worked examples:
Consumer API Spec Hackathon Review_Page 2- 1748631519807

FYI @stevespringett

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -97,20 +97,12 @@ paths:
         '404':
           $ref: "#/components/responses/404-object-by-id-not-found"
       tags:
-<<<<<<< Updated upstream
-        - TEA Component
-  /release/{uuid}/collection/latest:
-    get:
-      description: Get the latest TEA Collection belonging to the TEA Release
-      operationId: getLatestCollection
-=======
         - TEA Collection
   
   /component/{id}/collection/latest:
     get:
       description: Get the latest TEA Collection belonging for this TEA Component
       operationId: getLatestCollectionForComponent
->>>>>>> Stashed changes
       parameters:
         - name: id
           in: path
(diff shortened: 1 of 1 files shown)


ppkarwasz (changes requested):
Hi @madpah,

Thanks for sharing this! I think this PR has the potential to simplify TEA nicely. Even if it’s still an early draft, I wanted to highlight a few concerns and questions I have about the proposed changes.

taleodor:
@nscuro 
> I would suggest to not try to infer "latest" somehow - this is bound to fail due to the various ambiguities in versioning and release patterns. Instead, let's let publishers explicitly define what version is "latest". This mirrors what you can do in GitHub Releases, and FWIW, it's a mechanism that DT has adopted as well.

GitHub Releases has no concept of **latest**, I demonstrated this on one of the calls, can repeat today. Essentially, it may have releases from different streams all mashed together without clear order. Similarly, for DT, it's just a flag, but sometimes it's not meaningful. I.e., a project supports 2 major or minor versions (in semver) at the same time, each of them has their own latest.

> Possibly. Excuse my ignorance, and this is likely because I joined the effort so late, but where does the need to determine the "latest" component release stem from?

Component is an upstream dependency of Product. If there is a newer Component Release, as a developer / DevSecOps, I would like to migrate to that, especially if it has security updates. 

So the full example is like this - 
1. I'm using library ACMElib v.1.1.0
2. I want to know if there is a newer version available
3. ACMElib just released v.1.1.1 (which is a security patch for v.1.1.0), as well as v.1.2.0 (which is a new feature release).
4. I want TEA to inform me that v.1.1.1 is the one I can immediately upgrade to, while 1.2.0 is also there, but requires more testing since it's a feature update.

nscuro:
> In the practical sense of this PR, it means that if a user searches for TEI without version, this leads to a Product that lists all available versions and not to some latest version. And it should work the same way for Components. Products and Components are not that different.

Pragmatic, I like it.

> My understanding is that existing tooling should use publisher's TEA to make decisions rather than 3rd party heuristics.

If this is a goal then we should document it. I fear the `tea-usecases.md` doc may be horribly outdated at this point.

oej:
In the fall cleanup, I am going ahead and closing this. If there are things we missed that we still lack after all the recent changes, feel free to open issues or create a new PR.

Thank you for contributing!

PR #187, opened 2025-06-18
Update the discovery

oej:
- Add a note about avoiding guessing the TEI
- Add a note about "shared EAN codes" - those are not useful for creating unique TEIs for products that share EAN code. Each product needs a unique TEI.

### DIFF
--- a/discovery/readme.md
+++ b/discovery/readme.md
@@ -15,9 +15,10 @@
 ## From product identifier to API endpoint
 
 TEA Discovery is the connection between a product identifier and the API endpoint.
-A "product" is something that the customer aquires or downloads. It can be a bundle
-of many digital devices or software applications. A "product" normally also has an
-entry in a large corporation's asset inventory system.
+A "product" is something that the customer aquires or downloads - hardware and/or software.
+
+It can be a bundle of many digital devices or software applications.
+A "product" normally also has an entry in a large corporation's asset inventory system.
 
 A product identifier is embedded in a URN where the identifier is one of many existing
 identifiers or a random string - like an EAN or UPC bar code, UUID, product
--- a/discovery/readme.md
+++ b/discovery/readme.md
@@ -1,7 +1,5 @@
 # Transparency Exchange API - Discovery
 
-**NOTE**: _This is a proposal for the WG_
-
 - [From product identifier to API endpoint](#from-product-identifier-to-api-endpoint)
 - [Advertising the TEI](#advertising-the-tei)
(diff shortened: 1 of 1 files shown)


discovery/readme.md
```diff
@@ -48,20 +49,25 @@ and gives the vendor a name space to define product identifiers based on existin
...
+
 ## The TEI URN: An extensible identifier
 
 The TEI, Transparency Exchange Identifier, is a URN schema that is extensible based on existing
 identifiers like EAN codes, PURL and other identifiers. It is based on a DNS name, which leads
 to global uniqueness without new registries.
 
 The TEI can be shown in the software itself, in shipping documentation, in web pages and app stores.
-TEI is unique for a product, not a version of a software. The TEI consist of three core parts
+TEI is unique for a product, not a version of a product.
```
madpah, line 62: We also need to update the definition of Product in the beginning of this document as it does not cover physical (hardware) in my reading.

discovery/readme.md
```diff
@@ -149,6 +155,11 @@ urn:tei:uuid:cyclonedx.org:d4d9f54a-abcf-11ee-ac79-1a52914d44b1
 - GS1
 - STD
 
+Note that if an identifier, like EAN, is used for multiple different products then this
```
madpah, line 158: I think we should add SKU as an additional TEI type - e.g.:

`urn:tei:sku:tp-link.com:TL-SG2428P`

madpah: This would allow for regional variations of a Product to be handled, assuming they have separate SKUs within the Manufacturers domain:

e.g.:

`urn:tei:sku:tp-link.com:TL-SG2428P-CA`
`urn:tei:sku:tp-link.com:TL-SG2428P-EU`
`urn:tei:sku:tp-link.com:TL-SG2428P-US`

discovery/readme.md
```diff
@@ -48,20 +49,25 @@ and gives the vendor a name space to define product identifiers based on existin
...
 The TEI can be shown in the software itself, in shipping documentation, in web pages and app stores.
-TEI is unique for a product, not a version of a software. The TEI consist of three core parts
+TEI is unique for a product, not a version of a product.
 
 A TEI belongs to a single product. A product can have multiple TEIs - like one with a EAN/UPC
 barcode and one with the vendor's product number.
 
 ### TEI syntax
 
+The TEI consist of three core parts
```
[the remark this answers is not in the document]
oej: Fixed in 963ebde3ef4caded1bccea1ebaab907921fe6e28

PR #190, opened 2025-06-25
chore: GH workflow permissions

jkowalleck:


jkowalleck:
> _Nit_: wouldn't this be a good occasion to use `SHA1` to reference the exact version of GitHub actions we use?

Out of scope for this very pull requests 😁

PR #196, opened 2025-10-04
Streamline Discovery

taleodor:
This PR:

1. Converts ./well-known/tea endpoint to a static json object with defined schema.
2. Provides rules for discovery based on TEI following p.1
3. Adds error type to 404 responses, which can be OBJECT_UNKNOWN or OBJECT_NOT_SHAREABLE

### DIFF
--- a/discovery/readme.md
+++ b/discovery/readme.md
@@ -180,20 +180,65 @@ These point to the hosts available for the Transparency Exchange API.
 The TEA client connects to the host using HTTPS and validates the certificate.
 The URI is composed of the name with the `/.well-known/tea` prefix added.
 
-This results in the base URI (without the product identifier) 
-`https://tea.example.com/.well-known/tea/` 
+This results in the base URI 
+`https://tea.example.com/.well-known/tea`
+
+This URI must contain static json that lists the available TEA server endpoints and supported versions.
+The json must conform to the [TEA Well-Known Schema](tea-well-known.schema.json).
+
+Example:
+```json
+{
+  "endpoints": [
+    {
+      "url": "https://api.teaexample.com",
+      "versions": 
+        [
+          "0.1.0-beta.1",
+          "0.2.0-beta.2",
+          "1.0"
+        ]
+    },
+    {
+      "url": "https://api2.teaexample.com/mytea",
+      "versions": 
+        [
+          "1.0"
+        ]
+    }
+  ]
+}
+```
 
 
 ## Connecting to the API
 
-When connecting to the `.well-known/tea` URI with the unique identifier
-a HTTP redirect is **required**.
+Clients must pick any one of the endpoints listed in the `.well-known/tea` json
+response. They must then construct the full URL to the API by appending the
+"/v" plus one of the versions listed in the `versions` array of the selected endpoint,
+plus "/discovery?tei=", plus the TEI that is url-encoded according to [RFC3986]
+and [RFC3986]).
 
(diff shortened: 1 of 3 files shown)


oej:
This looks clever and may provide the functionality I was originally looking for. Earlier on there was a discussion that a .well-known URI was required to have a redirect, but from reading the RFC I don't believe that's the case any more.

Maybe we can add priority and/or weight like in SRV records to provide for load balancing and failover. Just a thought to consider.

The JSON structure in itself needs a version number so we can modify it in later versions of TEA. I don't think it needs to be the same as the TEA version number, but there may be benefits of using that. Either way, a version number of the data format will help as we move forward.

discovery/readme.md
```diff
@@ -180,35 +180,73 @@ These point to the hosts available for the Transparency Exchange API.
...
+```
 
 
 ## Connecting to the API
 
-When connecting to the `.well-known/tea` URI with the unique identifier
-a HTTP redirect is **required**.
+Clients must pick any one of the endpoints listed in the `.well-known/tea` json
+response. They must then construct the full URL to the API by appending the
+"/v" plus one of the versions listed in the `versions` array of the selected endpoint,
```
ppkarwasz, line 218: Does it make sense to append the **full** version of the API? By definition of backward compatibility a version `1.1` server **must** provide answers that are valid for a `1.0` client, even if it has an extended feature set (additional endpoints, additional query parameters and data in the responses).

taleodor: I like @oej suggestions, I'll try to implement them later this week (will add commits to this PR).

Regarding full version of the API, I think this is correct in our case. The reason being TEA is distributed in nature. Having only major versions works well when there is centralized API controlled by a single vendor. But in our case we may have client on version 1.5.0 and server on version 1.0.1 - making this /v1/ on the server would likely break such client.

An alternative approach is to use API version headers rather than include version in the path, but I think using the path is better (more explicit).

spec/openapi.yaml
```diff
@@ -378,6 +378,53 @@ paths:
...
+                properties:
+                  productReleaseUuid:
+                    description: UUID of the resolved TEA Product Release
+                    $ref: "#/components/schemas/uuid"
+                    example: d4d9f54a-abcf-11ee-ac79-1a52914d44b
+                  rootUrl:
+                    description: Root URL of the TEA server for this TEI without trailing slash
+                    type: string
+                    format: uri
+                    example: https://api.teaexample.com
```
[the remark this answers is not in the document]
taleodor: I think we should keep the discovery clean and straightforward. Having it potentially map to different types of objects would likely confuse clients and people. We may have separate endpoint(s) that do general search.

oej:
wikipedia on URN: "A [Uniform Resource Name](https://en.wikipedia.org/wiki/Uniform_Resource_Name) (URN) is a URI that identifies a resource by name in a particular namespace. A URN may be used to talk about a resource without implying its location or how to access it."

It's a name, not a location. Port numbers are location specific. One URN can be accessible in many servers on different ports.

oej:
I don't really dare proposing SRV again, but the new HTTPS records has a similar solution and they're standard. I don't think there's a lot of uptake on it though.

But yes, DNS is the way to go from a name to details. For the second part, we have your JSON structure which is almost DNS SRV, but I agree that discovery of the first part is missing. Maybe we can skip that for beta 2, get stuff out and come back to that later?

Issue #200, opened 2025-11-07
What productRelease do we return in `/discovery` if no version is provided


vpetersson:
I'm working on the internal data model for TEA in sbomify now and ran into a minor issue: 
For `/discovery`, what `productRelease` UUID do we return if no version is provided? In the case of a TEI that uses PURL, we can do some internal mapping to find the version, but this is not the case with say a GTIN.

My thinking is that re return the `productRelease` with the latest `createdDate`, bit this has it own set of challenges and might not necessarily return the right release.

I know @taleodor implemented the above flow using UUID [here](https://github.com/relizaio/rearm-cli/blob/main/docs/tea.md), but even for product UUID, we have the same version issue at play.

I am also fairly confident that we did discuss this in one of the meetings but I cannot recall the consensus.

taleodor:
The return type there is an array. The idea was to be able to return an array of matching product releases and then let the TEA client figure it out (which may be done with or without user help). Particularly, in my demo, the client was establishing a list discriminating components (those that use different releases between product releases) and then asking the user questions as to what component they use and then converge on the single product release based on the answers.

One more thing to add here - we discovered in the Hardware BOM group that sometimes it is literally impossible to distinguish between different product variants falling under the same identifier. We had similar examples in the TEA discussions as well (i.e., wifi routers with different hardware versioning but using same EAN/UPC). Hence the idea to allow to return an array of matching product releases.

Also note that the current wording in the discovery doc is the following:

"Note that if an identifier, like EAN, is used for multiple different product releases then this EAN code will not be unique for a given product. While this case is supported by TEA, the vendor is recommended to create a separate TEI for each unique product sold, like UUID or hash. In any case, the vendor SHOULD minimize the number of distinct product releases returned per TEI. Preferable situation is to have a single product release per TEI."

vpetersson:
> The return type there is an array. 

Ah, you're right. It is [defined as an array](https://github.com/CycloneDX/transparency-exchange-api/blob/main/spec/openapi.yaml#L1135). For whatever reason, Swagger UI is rendering it as a string.

In any case, that solves it for me.

Thanks!

vpetersson:
Wait, there is a bug in the example [here](https://github.com/CycloneDX/transparency-exchange-api/blob/main/spec/openapi.yaml#L1053). 

I assume:
```
example: d4d9f54a-abcf-11ee-ac79-1a52914d44b
```

Should be:
```
[...]

PR #201, opened 2025-11-07
Fixes discover UUID from string to array

vpetersson:
* Closes inconsistency between example and structure as discussed in https://github.com/CycloneDX/transparency-exchange-api/issues/200
* Clean up whitespace


vpetersson:
Thanks for the clarification. Closing out.

PR #215, opened 2026-02-22
Use distributionId instead of distributionType to identifiy disributions

### LINKED ISSUE #198: The distributionType field in release-distribution needs to be clarified
It's currently set as 

```
        distributionType:
          type: string
          description: Unique identifier for the distribution type.
```

Where examples include things like "zip", "tar.gz", "windows-x64.exe".

The problem - if we want to frame this as "unique identifier", this must be an enum. If we want to frame this as a free-form field describing packaging, that should be mentioned in the description and likely the field should be renamed. Right now, this is open to interpretation for publishers, and at the same time not clear how use that for clients - both of which things are not good.

taleodor:
This PR resolves #198 for the 1.0 release. It removes the distributionType string field as it is not clearly enumerated. Instead it introduces the distributionId UUID field that allows to map TEA Artifacts to specific distributions.

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -1030,6 +1030,7 @@ components:
       required:
         - uuid
         - type
+        - formats
     artifact-type:
       type: string
       description: Specifies the type of external reference.


oej (approved):
Looks good to me

Issue #172, opened 2025-05-28
Add `version` and `publishDate` to Artifact


madpah:
Use Case: VEX or VDR would regularly be updated. There are likely other scenarios too.

taleodor:
Also refer to #155 (this is partly duplicating).

oej:
Are we done here?

oej:
yes we are

PR #221, opened 2026-02-25
Add `date` field to TEA Artifact

ppkarwasz:
This change adds a `date` field to the TEA Artifact object.

Because TEA Artifacts are immutable, a single date field is sufficient; no distinction between multiple date types is required. The field name aligns with the existing `date` field in TEA Collection.

Conceptually, this field is comparable to `$.metadata.timestamp` in CycloneDX documents, which implementations may leverage.

Related to #172

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,10 +25,11 @@ informally using the Slack channel.
 
 
 Priority issues for v1.0:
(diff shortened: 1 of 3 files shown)


ppkarwasz:
Should we also rename the `date` field in a TEA Collection to make it more explicit? I can open a new issue for that.

Issue #179, opened 2025-05-28
Include Pagination across the Spec


taleodor:
Currently pagination is only added to /products endpoint, needs to be added to other relevant endpoints returning lists of elements.

nscuro:
I would still highly recommend to change the pagination API and not prescribe offsets. Some related background: https://use-the-index-luke.com/sql/partial-results/fetch-next-page

Opaque page tokens are more flexible and would allow the TEA implementation to use whatever mechanism they want behind the scenes. i.e. they could still encode offsets in the page tokens they generate. But this would not be a concern of TEA or its consumers.

taleodor:
I believe, based on the discussion during the last TC54-TG1 call on 2026-02-25, we want to have complete pagination system in place for v1.0 (not deferring).

Also, here are minimal tasks that I see as required regardless of the approach:
- Based on the current spec state, identify all endpoints that require pagination and current state of pagination (probably, AI can do it)
- Based on the comments above, it seems we need to prescribe canonical sorting for each such endpoint

Also, we may choose at this stage to drop offset based pagination entirely. We can do breaking changes before 1.0 and we don't have to preserve backward compatibility.

oej:
Agree. We need to get this major breaking change done before 1.0.

oej:
Merged

PR #231, opened 2026-03-12
feat: Add opaque pagination

madpah:
This PR (currently) adds Opaque Pagination to the `/products` endpoint as described in #179 - for review and comment before rolling it out to the other endpoints.

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -118,6 +118,10 @@ paths:
       description: Returns a list of TEA product releases. Note that multiple product releases may match.
       operationId: queryTeaProductReleases
       parameters:
+        # For result filtering
+        - $ref: "#/components/parameters/id-type"
+        - $ref: "#/components/parameters/id-value"
+        # For pagination
         - $ref: "#/components/parameters/page-size"
         - $ref: "#/components/parameters/page-token"
         - $ref: "#/components/parameters/sort-field-product-release"
@@ -159,6 +163,10 @@ paths:
         match.
       operationId: queryTeaProducts
       parameters:
+        # For result filtering
+        - $ref: "#/components/parameters/id-type"
+        - $ref: "#/components/parameters/id-value"
+        # For pagination
         - $ref: "#/components/parameters/page-size"
         - $ref: "#/components/parameters/page-token"
         - $ref: "#/components/parameters/sort-field-product"
@@ -247,6 +255,10 @@ paths:
       description: Returns a list of TEA components. Note that multiple components may match.
       operationId: queryTeaComponents
       parameters:
+        # For result filtering
+        - $ref: "#/components/parameters/id-type"
+        - $ref: "#/components/parameters/id-value"
+        # For pagination
         - $ref: "#/components/parameters/page-size"
         - $ref: "#/components/parameters/page-token"
         - $ref: "#/components/parameters/sort-field-component"
@@ -263,6 +275,10 @@ paths:
       description: Returns a list of TEA component releases. Note that multiple component releases may match.
       operationId: queryTeaComponentReleases
       parameters:
+        # For result filtering
+        - $ref: "#/components/parameters/id-type"
+        - $ref: "#/components/parameters/id-value"
+        # For pagination
         - $ref: "#/components/parameters/page-size"
         - $ref: "#/components/parameters/page-token"
         - $ref: "#/components/parameters/sort-field-component"
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -52,8 +52,10 @@ paths:
           description: UUID of TEA Product in the TEA server
           schema:
             "$ref": "#/components/schemas/uuid"
-        - $ref: "#/components/parameters/page-offset"
         - $ref: "#/components/parameters/page-size"
+        - $ref: "#/components/parameters/page-token"
+        - $ref: "#/components/parameters/sort-field-product-release"
+        - $ref: "#/components/parameters/sort-order"
       responses:
         '200':
           $ref: "#/components/responses/paginated-product-release"
@@ -116,10 +118,10 @@ paths:
       description: Returns a list of TEA product releases. Note that multiple product releases may match.
       operationId: queryTeaProductReleases
       parameters:
-        - $ref: "#/components/parameters/page-offset"
         - $ref: "#/components/parameters/page-size"
-        - $ref: "#/components/parameters/id-type"
-        - $ref: "#/components/parameters/id-value"
+        - $ref: "#/components/parameters/page-token"
+        - $ref: "#/components/parameters/sort-field-product-release"
+        - $ref: "#/components/parameters/sort-order"
       responses:
         '200':
           $ref: "#/components/responses/paginated-product-release"
@@ -203,15 +205,13 @@ paths:
           description: UUID of TEA Component in the TEA server
           schema:
             "$ref": "#/components/schemas/uuid"
+        - $ref: "#/components/parameters/page-size"
+        - $ref: "#/components/parameters/page-token"
+        - $ref: "#/components/parameters/sort-field-component-release"
+        - $ref: "#/components/parameters/sort-order"
       responses:
         '200':
-          description: Requested Releases of TEA Component found and returned
-          content:
-            application/json:
-              schema:
-                type: array
-                items:
-                  "$ref": "#/components/schemas/release"
+          $ref: "#/components/responses/paginated-component-release"
         '400':
           $ref: "#/components/responses/400-invalid-request"
         '404':
@@ -247,10 +247,10 @@ paths:
       description: Returns a list of TEA components. Note that multiple components may match.
       operationId: queryTeaComponents
       parameters:
-        - $ref: "#/components/parameters/page-offset"
         - $ref: "#/components/parameters/page-size"
-        - $ref: "#/components/parameters/id-type"
-        - $ref: "#/components/parameters/id-value"
+        - $ref: "#/components/parameters/page-token"
+        - $ref: "#/components/parameters/sort-field-component"
+        - $ref: "#/components/parameters/sort-order"
       responses:
         '200':
           $ref: "#/components/responses/paginated-component"
@@ -263,10 +263,10 @@ paths:
       description: Returns a list of TEA component releases. Note that multiple component releases may match.
       operationId: queryTeaComponentReleases
       parameters:
-        - $ref: "#/components/parameters/page-offset"
         - $ref: "#/components/parameters/page-size"
-        - $ref: "#/components/parameters/id-type"
-        - $ref: "#/components/parameters/id-value"
+        - $ref: "#/components/parameters/page-token"
+        - $ref: "#/components/parameters/sort-field-component"
+        - $ref: "#/components/parameters/sort-order"
       responses:
         '200':
           $ref: "#/components/responses/paginated-component-release"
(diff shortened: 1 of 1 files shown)


ppkarwasz (approved):

spec/openapi.yaml
```diff
@@ -157,10 +157,10 @@ paths:
         match.
       operationId: queryTeaProducts
       parameters:
-        - $ref: "#/components/parameters/page-offset"
         - $ref: "#/components/parameters/page-size"
-        - $ref: "#/components/parameters/id-type"
```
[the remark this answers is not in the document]
taleodor: Same here as my comment above -> id-type and id-value are not actually directly related to pagination.

taleodor (changes requested):
Hi @madpah , great work there.

I believe we should bring back all instances of id-type and id-value filters as those are not filters and not part of pagination. In relation to pagination, the server would need to take into account whether specific filtering was used. If that gets too complex, maybe we need a separate conversation on this.

Also we may need extra nesting due to behaviour of some generators, I'll try to test this over weekend based on your changes - essentially, at some point of TEA iterations, I was getting things like "inlinePaginationClass" without clear attribution to what class exactly was paginated. I'll get back on this but need a little extra time.

madpah:
> Hi @madpah , great work there.
> 
> I believe we should bring back all instances of id-type and id-value filters as those are not filters and not part of pagination. In relation to pagination, the server would need to take into account whether specific filtering was used. If that gets too complex, maybe we need a separate conversation on this.
> 
> Also we may need extra nesting due to behaviour of some generators, I'll try to test this over weekend based on your changes - essentially, at some point of TEA iterations, I was getting things like "inlinePaginationClass" without clear attribution to what class exactly was paginated. I'll get back on this but need a little extra time.

Understood - the purpose of those fields was not immediately clear - so I'll read add and aim to make it clear they are about result filtering and NOT pagination.

spec/openapi.yaml
```diff
@@ -116,10 +118,10 @@ paths:
       description: Returns a list of TEA product releases. Note that multiple product releases may match.
       operationId: queryTeaProductReleases
       parameters:
-        - $ref: "#/components/parameters/page-offset"
         - $ref: "#/components/parameters/page-size"
-        - $ref: "#/components/parameters/id-type"
```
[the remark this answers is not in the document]
madpah: Re-added @taleodor.

spec/openapi.yaml
```diff
@@ -263,10 +263,10 @@ paths:
       description: Returns a list of TEA component releases. Note that multiple component releases may match.
       operationId: queryTeaComponentReleases
       parameters:
-        - $ref: "#/components/parameters/page-offset"
         - $ref: "#/components/parameters/page-size"
-        - $ref: "#/components/parameters/id-type"
-        - $ref: "#/components/parameters/id-value"
```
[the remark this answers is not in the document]
madpah: Re-added @taleodor.

madpah:
> Hi @madpah , great work there.
> 
> I believe we should bring back all instances of id-type and id-value filters as those are not filters and not part of pagination. In relation to pagination, the server would need to take into account whether specific filtering was used. If that gets too complex, maybe we need a separate conversation on this.
> 
> Also we may need extra nesting due to behaviour of some generators, I'll try to test this over weekend based on your changes - essentially, at some point of TEA iterations, I was getting things like "inlinePaginationClass" without clear attribution to what class exactly was paginated. I'll get back on this but need a little extra time.

FWIW - the generated libs from this schema seem fine and work for me - if you see issues, let me know - might require a fix upstream in OpenAPI Generator 😉

madpah:
What's the issue with the TeaInline* objects being generated @taleodor? This is fine in other languages for sure - so not sure of the issue here.

madpah:
Yeah - I see the inline objects - but don't understand what the issue is here @taleodor - is it because the naming is not deterministic, or something else?

taleodor:
> Yeah - I see the inline objects - but don't understand what the issue is here @taleodor - is it because the naming is not deterministic, or something else?

Yes, essentially we can't build proper server code like that. Like we need to use these classes to build pagination but we can't do it in a meaningful way - it's like working with obfuscated code.

madpah:
Not sure I agree with the assessment @taleodor - I've used this before in both generated server and client code. The only real issue I've experienced is non-deterministic names of Classes which can be a pain.

Do you have a repo or commit somewhere that illustrates your pain that we can reference?

taleodor:
I believe we shouldn't leave it like that in any case since it's public spec and this creates a lot of friction and uncertainty on the implementation side. I'll propose a fix to this PR - should have time over the weekend.

taleodor:
@madpah I opened #243 against your PR branch which should fix this.

spec/openapi.yaml
```diff
@@ -1568,45 +1530,168 @@ components:
...
-      name: pageSize
-      description: Pagination offset
+        minimum: 1
+        maximum: 100
+        default: 25
+    page-token:
+      name: pageToken
+      description: |
+        An opaque token used to retrieve the next page of results. 
+        This should be copied exactly from the `nextPageToken` field of a previous response.
```
Mehrn0ush, line 1623: Can we define how `pageToken` interacts with the other query parameters?

For cursor/token pagination, clients need to know whether `pageToken` already encapsulates the original query state, including filters, `sortField`, `sortOrder`, and possibly `pageSize`.

For example, if the first request uses:

`sortField=createdDate&sortOrder=desc&pageSize=25`

and the next request sends:

`pageToken=...&sortField=version&sortOrder=asc&pageSize=100`

what should the server do?

Possible options:

1. The server MUST ignore `sortField`, `sortOrder`, filters, and `pageSize` when `pageToken` is present.
2. The server MUST return `400` if parameters conflict with the state encoded in `pageToken`.
3. The client MUST repeat the same parameters, and mismatch is an invalid request.

I think the spec should choose one behavior before v1.0 to avoid incompatible implementations. Defining this behavior now will ensure forward compatibility and consistent behavior across different TEA server implementations.

spec/openapi.yaml
```diff
@@ -1468,74 +1480,24 @@ components:
...
-        - $ref: "#/components/schemas/pagination-details"
-        - type: object
-          properties:
-            results:
-              type: array
-              items:
-                $ref: "#/components/schemas/component"
+            It must always be supplied in responses.
+      required:
+        - hasNext
```
Mehrn0ush, line 1498: Could we clarify the expected semantics of `nextPageToken` when `hasNext` is `false`?

Right now `nextPageToken` is required, non-nullable, and described as always supplied. If there is no next page, it is unclear whether the server should return an empty string, a terminal token, or some other value.

Would it be cleaner to make `nextPageToken` optional and present only when `hasNext=true`?

For example:

- `hasNext=true`: `nextPageToken` MUST be present.
- `hasNext=false`: `nextPageToken` MUST be absent.

Alternatively, if the intention is to always return a checkpoint token for future polling/sync, maybe that should be modeled separately from a next-page token, because those are slightly different semantics.

spec/openapi.yaml
```diff
@@ -1468,74 +1480,24 @@ components:
     pagination-details:
       type: object
       properties:
-        timestamp:
+        hasNext:
+          type: boolean
+          description: |
+            A flag (to aid clients) to know whether there is a next page of results to fetch. 
+            
+            `nextPageToken` will always be supplied, hence this hint is included to aid clients.
```
Mehrn0ush, line 1488: The `hasNext` description currently says that `nextPageToken` will always be supplied.

If we decide that `nextPageToken` should only be present when `hasNext=true`, this description should be updated accordingly. Otherwise, the spec should clarify what value `nextPageToken` has when `hasNext=false`. 
Making nextPageToken optional when hasNext is false would simplify client-side logic, as consumers wouldn't need to handle 'empty' or 'dummy' tokens for terminal pages.

spec/openapi.yaml
```diff
@@ -407,15 +421,13 @@ paths:
           description: UUID of TEA Product Release in the TEA server
           schema:
             "$ref": "#/components/schemas/uuid"
+        - $ref: "#/components/parameters/page-size"
+        - $ref: "#/components/parameters/page-token"
+        - $ref: "#/components/parameters/sort-field-collection"
+        - $ref: "#/components/parameters/sort-order"
```
Mehrn0ush, line 427: Could we define a deterministic tie-breaker for each sortable resource type?

The addition of `sortField` is helpful, but fields like `createdDate`, `releaseDate`, `version`, and `name` are not guaranteed to be unique. Without a stable tie-breaker, cursor pagination can still produce duplicate or missing results when multiple records share the same sort value.

A simple rule could be:

`Results MUST be ordered by the selected sortField, then by UUID as a deterministic tie-breaker.`

This would make token pagination more reliable and easier to implement consistently.

spec/openapi.yaml
```diff
@@ -381,15 +397,13 @@ paths:
           description: UUID of TEA Component Release in the TEA server
           schema:
             "$ref": "#/components/schemas/uuid"
+        - $ref: "#/components/parameters/page-size"
+        - $ref: "#/components/parameters/page-token"
+        - $ref: "#/components/parameters/sort-field-collection"
+        - $ref: "#/components/parameters/sort-order"
```
Mehrn0ush, line 403: Could we define a deterministic tie-breaker for each sortable resource type?

The addition of `sortField` is helpful, but fields like `createdDate`, `releaseDate`, `version`, and `name` are not guaranteed to be unique. Without a stable tie-breaker, cursor pagination can still produce duplicate or missing results when multiple records share the same sort value.

A simple rule could be:

`Results MUST be ordered by the selected sortField, then by UUID as a deterministic tie-breaker.`

This would make token pagination more reliable and easier to implement consistently.

spec/openapi.yaml
```diff
@@ -1568,45 +1530,168 @@ components:
...
+        default: createdDate
+    sort-field-product:
+      name: sortField
+      description: The field by which to sort the results.
+      in: query
+      required: false
+      schema:
+        type: string
+        enum:
+          - name
```
Mehrn0ush, line 1680: sort-field-component:
  enum:
    - name 
   
and 

sort-field-product:
  enum:
    - name

If `name` is the default sort field for products and components, do we also need to specify whether names are unique within a TEA server?

If names are not guaranteed to be unique, token pagination still needs a deterministic secondary ordering, such as UUID. If names are intended to be unique, that constraint should probably be documented in the corresponding product/component schema descriptions.

spec/openapi.yaml
```diff
@@ -1568,45 +1530,168 @@ components:
...
+      name: sortOrder
+      description: The direction of the sort.
+      in: query
+      required: false
+      schema:
+        type: string
+        enum:
+          - asc
+          - desc
+        default: asc
```
Mehrn0ush, line 1638: I noticed sortOrder defaults to asc globally. Would it be worth considering resource-specific defaults instead?

For security-related resources like product/component releases (often sorted by createdDate) or collections (sorted by version), consumers typically expect the most recent data first (newest-to-oldest).

If a universal asc default is preferred for simplicity, perhaps we could add a small note in the description to help set expectations for implementers/consumers.

oej:
We've decided to move unresolved comments to new issues and merge this as is, to be improved in separate pull requests.

spec/openapi.yaml
oej: Moved to #247

spec/openapi.yaml
oej: Moved to #248

spec/openapi.yaml
oej: Moved to #249

spec/openapi.yaml
oej: Moved to #250

spec/openapi.yaml
oej: Moved to #251

PR #235, opened 2026-03-29
spec: add TEA discovery, protobuf, and JSON schema drafts

MChorfa:
## Summary
This PR carries forward only the specification-facing parts of the previous `ref-impl` branch.

Included here:
- discovery and API flow documentation
- TEA protobuf definitions under `proto/tea/v1`
- JSON schemas under `schemas/json`
- supporting spec-oriented diagrams and reference docs

Explicitly excluded from this PR:
- Rust server implementation
- Dagger/build automation
- CI/release/supply-chain workflows
- generated integration/reference-implementation material

## Implementation split
The implementation work has been moved into a separate repository:
- https://github.com/MChorfa/transparency-exchange-reference-implementation

## Notes
This is intended to align with the repository boundary discussed in the previous review and keep this PR limited to spec/schema/doc content.

PR #236, opened 2026-03-30
doc: narrow TEA authentication guidance

MChorfa:
## Summary
This PR is a focused split from #235 and contains only `doc/authentication.md`.

## What changed
- removes hard-coded universal claims such as a mandatory TLS 1.3 minimum and specific client certificate algorithms
- describes bearer tokens and mTLS as common deployment patterns
- states that stricter TLS baselines belong in a TEA profile or deployment baseline
- references OWASP TLS guidance and NIST SP 800-52 Rev. 2

## Intent
Address review feedback that the prior TLS language was too broad and not grounded enough for the base TEA specification.


MChorfa:
Following the feedback on #235: the earlier TLS wording was carried forward from the broader draft rather than intended as a final universal base-spec requirement, and the concern about it being too prescriptive was fair.

This follow-up narrows the document to deployment patterns, keeps only the base expectation that protected TEA endpoints use TLS, and leaves stricter transport baselines to TEA profiles or deployment policy.

oej:
Is a new document needed? What is missing in the current documentation?

oej:
Closing due to no response from contributor.

PR #253, opened 2026-06-03
Clarify pagination continuation and sorting rules

### LINKED ISSUE #249: Could we define a deterministic tie-breaker for each sortable resource type?
Could we define a deterministic tie-breaker for each sortable resource type?

The addition of `sortField` is helpful, but fields like `createdDate`, `releaseDate`, `version`, and `name` are not guaranteed to be unique. Without a stable tie-breaker, cursor pagination can still produce duplicate or missing results when multiple records share the same sort value.

A simple rule could be:

`Results MUST be ordered by the selected sortField, then by UUID as a deterministic tie-breaker.`

This would make token pagination more reliable and easier to implement consistently.

_Originally posted by @Mehrn0ush in https://github.com/CycloneDX/transparency-exchange-api/pull/231#discussion_r3186357701_

### LINKED ISSUE #250: The `hasNext` description currently says that `nextPageToken` will always be supplied.
The `hasNext` description currently says that `nextPageToken` will always be supplied.

If we decide that `nextPageToken` should only be present when `hasNext=true`, this description should be updated accordingly. Otherwise, the spec should clarify what value `nextPageToken` has when `hasNext=false`. 
Making nextPageToken optional when hasNext is false would simplify client-side logic, as consumers wouldn't need to handle 'empty' or 'dummy' tokens for terminal pages.

_Originally posted by @Mehrn0ush in https://github.com/CycloneDX/transparency-exchange-api/pull/231#discussion_r3186356724_

### LINKED ISSUE #251: Can we define how `pageToken` interacts with the other query parameters?
Can we define how `pageToken` interacts with the other query parameters?

For cursor/token pagination, clients need to know whether `pageToken` already encapsulates the original query state, including filters, `sortField`, `sortOrder`, and possibly `pageSize`.

For example, if the first request uses:

`sortField=createdDate&sortOrder=desc&pageSize=25`

and the next request sends:

`pageToken=...&sortField=version&sortOrder=asc&pageSize=100`

what should the server do?

Possible options:

1. The server MUST ignore `sortField`, `sortOrder`, filters, and `pageSize` when `pageToken` is present.
2. The server MUST return `400` if parameters conflict with the state encoded in `pageToken`.
3. The client MUST repeat the same parameters, and mismatch is an invalid request.

I think the spec should choose one behavior before v1.0 to avoid incompatible implementations. Defining this behavior now will ensure forward compatibility and consistent behavior across different TEA server implementations.

_Originally posted by @Mehrn0ush in https://github.com/CycloneDX/transparency-exchange-api/pull/231#discussion_r3186355852_

Mehrn0ush:
This PR clarifies the opaque pagination contract in `spec/openapi.yaml` and addresses issues #249, #250, and #251.

The changes define deterministic ordering rules, clarify `hasNext` / `nextPageToken` behavior, and specify how `pageToken` interacts with query parameters and path-scoped resources.

### Deterministic pagination ordering (#249)

- Define deterministic ordering for paginated results.
- Product, component, product release, and component release results are ordered first by the selected `sortField`, then by `uuid` as a deterministic secondary tie-breaker.
- Collection pagination uses collection-specific ordering and does not rely on `uuid` because collection revisions may share the same UUID.
- Clarify release `version` sorting semantics and explicitly state that semantic-version precedence is not implied.
- Define ordering behavior for missing `releaseDate` values.
- Fix `/componentReleases` to use `sort-field-component-release` instead of `sort-field-component`.

### `hasNext` and `nextPageToken` (#250)

- Remove `nextPageToken` from required pagination response fields.
- Specify that `nextPageToken` MUST be present when `hasNext` is `true`.
- Specify that `nextPageToken` MUST NOT be included when `hasNext` is `false`.
- Add terminal and non-terminal pagination examples.

### `pageToken` continuation semantics (#251)

- Clarify that `pageToken` is opaque and MUST be copied verbatim from a previous `nextPageToken`.
- Specify that clients MUST NOT parse, construct, or modify `pageToken`.
- Define continuation state, including:
  - `pageSize`
  - `sortField`
  - `sortOrder`
  - result-affecting filters such as `idType` and `idValue`
  - request path and path parameters such as parent `uuid`
- Specify that clients MUST NOT change result-affecting parameters while using a continuation token.
- Specify that servers MUST return `400 Bad Request` for malformed, invalid, expired, conflicting, wrong-path, or wrong-path-parameter tokens.

Closes #249  
Closes #250  
Closes #251

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -281,7 +281,7 @@ paths:
         # For pagination
         - $ref: "#/components/parameters/page-size"
         - $ref: "#/components/parameters/page-token"
-        - $ref: "#/components/parameters/sort-field-component"
+        - $ref: "#/components/parameters/sort-field-component-release"
         - $ref: "#/components/parameters/sort-order"
       responses:
         '200':
@@ -1483,20 +1483,24 @@ components:
         hasNext:
           type: boolean
           description: |
-            A flag (to aid clients) to know whether there is a next page of results to fetch. 
-            
-            `nextPageToken` will always be supplied, hence this hint is included to aid clients.
+            A flag indicating whether there is another page of results to fetch.
+
+            When `hasNext` is `true`, `nextPageToken` MUST be present.
+            When `hasNext` is `false`, `nextPageToken` MUST NOT be included.
           default: false
         nextPageToken:
           type: string
-          nullable: false
           description: |
-            A token that can be used in a following request to retrieve the next page or results.
+            An opaque token that can be used in a following request to retrieve the next page of results.
 
-            It must always be supplied in responses.
+            This field MUST be present when `hasNext` is `true`.
+            This field MUST NOT be included when `hasNext` is `false`.
       required:
         - hasNext
-        - nextPageToken
+      examples:
+        - hasNext: true
+          nextPageToken: eyJwYWdlIjoiY29udGludWF0aW9uIn0
+        - hasNext: false
     paginated-product-response:
       description: A paginated response containing TEA Products
       allOf:
@@ -1509,6 +1513,12 @@ components:
                 $ref: "#/components/schemas/product"
           required:
             - results
+      examples:
+        - hasNext: true
+          nextPageToken: eyJwYWdlIjoiY29udGludWF0aW9uIn0
+          results: []
+        - hasNext: false
+          results: []
     paginated-product-release-response:
       description: A paginated response containing TEA Product Releases
       allOf:
@@ -1521,6 +1531,12 @@ components:
                 $ref: "#/components/schemas/productRelease"
           required:
             - results
+      examples:
+        - hasNext: true
+          nextPageToken: eyJwYWdlIjoiY29udGludWF0aW9uIn0
+          results: []
+        - hasNext: false
+          results: []
     paginated-component-response:
       description: A paginated response containing TEA Components
       allOf:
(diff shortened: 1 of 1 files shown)


spec/openapi.yaml
```diff
@@ -1632,15 +1669,38 @@ components:
     page-token:
       name: pageToken
       description: |
-        An opaque token used to retrieve the next page of results. 
-        This should be copied exactly from the `nextPageToken` field of a previous response.
+        An opaque continuation token produced by a previous response.
+        This token MUST be copied verbatim from the `nextPageToken` value returned by the previous response.
+        Clients MUST NOT parse, construct, or modify this token.
+
+        The token represents continuation state for the original query, including `pageSize`,
```
taleodor: I would suggest excluding pageSize from this list, since that is not affecting results, and can be left to clients.

taleodor (changes requested):
Great points, just want to request one change to exclude pageSize from the token scope.

spec/openapi.yaml
Mehrn0ush: Thanks, that makes sense. I updated the `pageToken` description to remove `pageSize` from the continuation state. The token scope now remains limited to sorting, result-affecting filters, and path/path-parameter scope.

taleodor (approved):

oej (approved):
Discussed in meeting Aug 12 '26

- Approved by Pavel and Viktor

Decision on merging! Thank you

PR #254, opened 2026-08-03
docs(implementations): add Vulnetix entries and move the list to tables

0x73746F66:
## What

Adds two Vulnetix implementations to `doc/tea-implementations.md` and reorganises the page into tables.

**New entries**

- **Vulnetix CLI and GitHub Action** (open source clients) AGPL-3.0 golang cli acting as both a TEA v0.4.0 consumer and producer. [Docs](https://docs.cli.vulnetix.com/docs/cli-reference/tea/) · [GitHub](https://github.com/Vulnetix/cli)
- **Vulnetix TEA Explorer** (other implementations) — a TEA v0.4.0 compatible GUI with TEA discovery that shows the API call behind every page of data. [Product details](https://www.vulnetix.com/features/transparency-exchange)

**Reorganisation**

Each category is now a single table with `Implementation`, `Type`, `Description` and `Resources` columns. Existing copy and links are carried over unchanged, so nothing is lost, and the entries become comparable at a glance as the list grows.

## Why

The list was a flat run of headings with no consistent shape, so a reader could not tell at a glance whether an entry was a client, a server, a library or a GUI, nor which TEA version it targeted. The table makes that explicit and gives future submissions a fixed format to follow.

### DIFF
--- a/doc/tea-implementations.md
+++ b/doc/tea-implementations.md
@@ -2,68 +2,29 @@
 
 # TEA Implementations
 
-
 ## Open Source clients
 
-### py-libtea
-
-A Python client library and CLI client for TEA.
-
-- [GitHub](https://github.com/sbomify/py-libtea)
-
-### PyPi TEA
-
-A TEA bridge for accessing Python SBOMs from PyPi (via [PEP 770](https://peps.python.org/pep-0770/)).
-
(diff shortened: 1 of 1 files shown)


oej (approved):
THanks

PR #257, opened 2026-08-26
Token Auth Endpoint

taleodor:


taleodor:
recreating yet again due to gitguardian

oej:
I do not think we should support "HTTP basic". Only "HTTP Digest" if we need one. Refer to RFC 7616.

PR #258, opened 2026-08-26
feat: token exchange endpoint and documentation

taleodor:

### DIFF
--- a/auth/readme.md
+++ b/auth/readme.md
@@ -28,7 +28,11 @@ authorization - does not make much sense, since the information is usually in th
 ## Scope of this specification
 
 This specification does not require a TEA service to authenticate its users. A service that
-publishes openly need not implement any of what follows.
+publishes openly need not implement any of what follows, including the token endpoint; see
+[Servers without authentication](#servers-without-authentication) for what such a server and its
+clients do instead. A service that requires authentication for any of its endpoints is a service
+that requires authentication, and implements the baseline in full; see
+[Mixed servers](#mixed-servers).
 
 Where a service does authenticate, interoperability requires that every TEA client can
 authenticate against every TEA server without server-specific code. This specification therefore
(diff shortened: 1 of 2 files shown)


auth/readme.md
```diff
@@ -25,36 +25,205 @@ the customer has not aquired.
...
+  key directly on the resource endpoints, and a client __shall not__ present one there. The API key
+  is exchanged for an access token, and the access token is what the resource endpoints see.
+* Whatever credential a client starts with, it ends up holding the same thing: a TEA access token
+  presented as an HTTP bearer token. Clients therefore need one code path for the API itself,
+  regardless of how the server manages identity.
+
+## The token endpoint
+
+The token endpoint is `POST /token`, relative to the TEA API base URL, and is defined in
+[the OpenAPI specification](../spec/openapi.yaml). It is an OAuth 2.0 token endpoint as defined
```
oej: Will this link survive the official doc process (ECMA docs) ?

auth/readme.md
```diff
@@ -25,36 +25,205 @@ the customer has not aquired.
...
+__The access token is opaque to the client.__ Clients __shall not__ inspect, parse, or depend on the
+contents of the token, and __shall not__ assume any particular format. Servers are free to issue a
+random handle, a signed token such as a JWT access token
+([RFC 9068](https://www.rfc-editor.org/rfc/rfc9068)), or anything else, and to change that choice
+without notice; only the issuing server interprets it.
+
+Token lifetime, storage, rotation, and revocation are implementation matters and are deliberately
+not specified. Two rules keep clients simple in spite of that:
+
+* Servers __should__ return `expires_in`, so that a client can obtain a new token before the current
```
oej, line 136: Is this a header, part of the JWT or?

oej (changes requested):
Good work, just some smaller questions

oej:
If a client connects to a server that has no "/token" endpoint and no auth - will it receive a 404 which feels weird.

i would like that *every* TEA service has /token endpoint and if no auth is in play, simply return "200 ok"

auth/readme.md
taleodor: Good point, text amended accordingly.

auth/readme.md
taleodor: This is a field on token response, described in the openapi spec:

        expires_in:
          type: integer
          description: |
            Lifetime of the access token in seconds. Servers should include this so that
            clients can re-authenticate before expiry rather than on failure.
          example: 3600

taleodor:
> If a client connects to a server that has no "/token" endpoint and no auth - will it receive a 404 which feels weird.
> 
> i would like that _every_ TEA service has /token endpoint and if no auth is in play, simply return "200 ok"

@oej Clarified this in follow up commit. Essentially, client never starts with accessing token endpoint, it only does so if receives 401 prior. So client shouldn't be probing /token.

oej (approved):
Approved in meeting

PR #260, opened 2026-08-26
Align markdown docs with OpenAPI specfication

oej:
- Remove duplicate text
- Move text that's ended up in the wrong doc


taleodor:
Some inconsistencies in this PR:

- The `date` → `createdDate` rename on the collection hits the schema only. Both spec examples still say `date`, and the collection markdown's JSON example does too, while its field list now says `createdDate`. #264 makes the identical rename, so the two PRs conflict on the same hunk.
- The moved release documentation uses `distributionType` throughout its prose and JSON examples, while the new "key attributes" list correctly says `distributionId`, which is the schema property. The examples lack the required `distributionId`, so they fail the schema the PR aligns with. The artifact section of the collection doc has the same split.
- The pre-release text was moved into the product-release doc verbatim from the component doc, so the product-release page now explains pre-releases in terms of "creating the Component with a new UUID". The release doc gets a different, divergent description of the same flag.
- Spec examples switch to `tei://`, but the component doc's new sentence still says "TEI URNs", and the product-release markdown example still uses a third form, `tei:vendor:product@2.24.3`.
- The `date-time` description now claims RFC 9557, while the pattern on the next line only admits `YYYY-MM-DDTHH:MM:SSZ`. The description and the regex disagree.
- Moved checksum examples use `SHA_256` and `SHA_512` with underscores. The enum is hyphenated, so those examples fail validation. Pre-existing, but the PR copies them forward as the aligned version.
- The discovery example ends with an encoded newline, `%0A`, which is an invalid value regardless of the URN-versus-URL decision.

**Overlap with #262**

This PR rewrites the `artifact-format` `url` / `signatureUrl` descriptions in the spec and the matching `mediaType` / `url` / `signatureUrl` bullets in `tea-collection.md`, which are exactly the lines #262 changes (`mediaType` required; `url` and `signatureUrl` always external and optional, with the server's download endpoints used when absent). Here `url` is left unmarked next to `signatureUrl (optional)`, which reads as required, so it's a semantic disagreement as well as a merge conflict. Suggest rebasing this on #262 once it lands and taking those bullets from there, so the docs end up consistent with the download endpoints.

oej:
Fixed some comments, waiting for #262 for another review

oej:
Rebased to current main branch, resolved conflicts

Mehrn0ush:
Three schema-alignment :

1. tea-release.md examples/prose still use distributionId values like "jar" / "zip"; the schema requires UUIDs.
2. tea-collection.md still documents SHA_256/SHA3_512, ships an MD5 example, and mentions distributionType instead of distributionIds; artifact examples should include required createdDate (and version where applicable).
3. Product-release example still has tei:vendor:product@2.24.3 instead of a tei://… URL.

PR #262, opened 2026-08-26
feat: artifact and signature download endpoints

taleodor:
Adds endpoints to download TEA Artifact and signature contents.

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -1346,7 +1346,7 @@ components:
           description: |
             The Media Type of the document.
 
-            A media type MUST appear at most once across the `formats` of a single TEA
+            A media type Shall appear at most once across the `formats` of a single TEA
             Artifact revision, so that a format can be selected unambiguously by media
             type when downloading its content.
         description:
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -548,6 +548,160 @@ paths:
           $ref: "#/components/responses/404-object-by-id-not-found"
       tags:
         - TEA Artifact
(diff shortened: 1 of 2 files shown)


Mehrn0ush:
Hi,

Following up from the meeting — `latest/download` and `{artifactVersion}/download` currently share the same artifact-content response component, which means they both get `Cache-Control: public, max-age=31536000`, immutable. That's fine for the versioned endpoint since a revision never changes, but not for latest, since that's exactly the endpoint whose target does change over time. Right now a cache could hang onto a year-old copy and never revalidate, even after a new revision is published. Same problem on the signature side (`latest/signature/download` / `artifact-signature-content`).
I think latest/* needs its own cache policy, no immutable, probably no-cache or a short max-age instead.

Mehrn0ush:
One more question:
`artifact-format.mediaType` —  mentions a media type MUST appear at most once across a revision's formats, which is good, but mediaType itself isn't marked required on the schema. If a format can be published without one, then selecting a download by mediaType query param or by Accept has nothing to match against for that format — how should a client target it, and how does the "at most once" uniqueness rule even apply to an absent value?

oej:
> One more question: `artifact-format.mediaType` — mentions a media type MUST appear at most once across a revision's formats, which is good, but mediaType itself isn't marked required on the schema. 

I think mediaType should be required.

taleodor:
@Mehrn0ush @oej I believe those points above are addressed now, pls check.

Mehrn0ush:
Thanks @taleodor 
Looks good. latest vs versioned cache split and required mediaType address the earlier comments.

PR #264, opened 2026-08-31
Openapi require changes

oej:
Adding a set of required fields.

Closes issue #259 

Please review!


spec/openapi.yaml
```diff
@@ -1022,7 +1036,7 @@ components:
           description: |
             TEA Collection version, incremented each time its content changes.
             Versions start with 1.
-        date:
+        createdDate:
```
taleodor, line 1039: This should be updated in examples that still use `date`.

spec/openapi.yaml
```diff
@@ -1054,8 +1074,6 @@ components:
                   url: https://repo.maven.apache.org/maven2/org/apache/logging/log4j/log4j-core/2.24.3/log4j-core-2.24.3-cyclonedx.xml
                   signatureUrl: https://repo.maven.apache.org/maven2/org/apache/logging/log4j/log4j-core/2.24.3/log4j-core-2.24.3-cyclonedx.xml.asc
                   checksums:
-                    - algType: MD5
```
taleodor, line 1057: This should be reflected in examples that still use MD5.

taleodor:
Added comments, plus:

New required fields not present in examples: product on the product release, component on the component release, belongsTo on the collection, and createdDate on artifacts are absent from the existing examples, so every one of those examples fails the schema it sits under.

taleodor (dismissed):

Mehrn0ush:
I had two questions about modeling that appeared worth clarifying during this tightening process.

1. Should a `productRelease` be allowed to exist without an associated product?

Requiring product makes sense if TEA 1.0 intends every release to resolve to a registered Product.

The alternative would be to keep it optional for cases where someone publishes release-level artifacts such as an SBOM or VEX before registering a full Product entity, or where a release may never have one.

If every release should always belong to a Product, the new requirement looks right and the related docs can be brought into sync. If not, product should probably move back out of required.

2. Should artifacts be required on collection, while still allowing `artifacts: []`?

I’d lean toward yes. Requiring the field would distinguish an explicitly empty collection from a missing/unknown value, and it would also simplify generated clients by avoiding Artifact[] | undefined.

The main downside I can see is that it removes the ability to use omission to mean “unknown/not disclosed.” If TEA doesn’t need that distinction, a required-but-empty array seems cleaner.

Related to that, since `artifact.formats` is already required, I think `minItems: 1` would also make sense so an artifact can’t exist without at least one downloadable representation.

oej:
Good questions. I think we have to require a base product to register a product release.

A collection for a component release that has no artifacts is kind of strange, but with your additional input I think it should be allowed.

oej:
Updated

Mehrn0ush:
Thanks. 

Two small leftovers:

README still says a Product Release may optionally belong to a TEA Product, but product is now required on productRelease. Worth aligning the README (and related tea-product docs, which has similar wording) with that model.

Formats is already required, would it make sense to add `minItems: 1` so an artifact can't advertise formats: [] with no downloadable representation?

Typo: a few checksum examples still use SHA_256/SHA_512 (underscore) instead of the enum's hyphenated form (SHA-256, SHA-512) — looks like it's fixed in some spots but not all.

oej:
- Fixed so that docs should be more aligned with spec requirement of product in product release
- Fixed "SHA_"
- Added "minItems" on formats

Mehrn0ush:
Thanks
A few leftovers:

1. README still says optional product: (data-model bullets): "may optionally belong to a TEA Product" / Product as "optional higher-level object". That contradicts required `product` on `productRelease`.
2. First `artifact` example is missing required `version` (second example has it).
3. Pre-release Tomcat `release` example: (`11.0.0-M26`) is missing required `component`.
4. Checksum after MD5 removal: `algType: SHA-256` with a 32-hex `algValue` (old MD5 length). Needs a real SHA-256 hex (64 chars) or drop that entry.

oej:
> Thanks A few leftovers:
> 
> 1. README still says optional product: (data-model bullets): "may optionally belong to a TEA Product" / Product as "optional higher-level object". That contradicts required `product` on `productRelease`.
> 2. First `artifact` example is missing required `version` (second example has it).
> 3. Pre-release Tomcat `release` example: (`11.0.0-M26`) is missing required `component`.
> 4. Checksum after MD5 removal: `algType: SHA-256` with a 32-hex `algValue` (old MD5 length). Needs a real SHA-256 hex (64 chars) or drop that entry.

Fixed in recent commits. Please review again.

Mehrn0ush:
Thanks 

LGTM from my side on those points.

taleodor:
Three things left.

**1. `README.md` line 58 still contradicts the model.** It describes TEA Product as "An **optional** higher-level object that groups a set of Product Releases", while `product` is now required on `productRelease`. `tea-product.md` and `tea-product-release.md` were aligned; the README was not.

**2. `collection.examples[0]` — both artifacts fail `artifact.required`** (`[uuid, type, formats, createdDate, version]`):

- `artifacts[0]` has only `formats, name, type, uuid` — missing `createdDate` and `version`
- `artifacts[1]` has `formats, name, type, uuid, version` — missing `createdDate`

**3. `component-release-with-collection.examples[0]` — three failures:**

- the embedded `release` has `createdDate, identifiers, releaseDate, uuid, version` — missing required `component`. This is the Tomcat **11.0.7** example; `7bdfff0` added `component` to the pre-release `11.0.0-M26` example but not to this one.
- `latestCollection.artifacts[0]` and `latestCollection.artifacts[1]` — both missing `createdDate` and `version`

oej:
> Three things left.
> 
> **1. `README.md` line 58 still contradicts the model.** It describes TEA Product as "An **optional** higher-level object that groups a set of Product Releases", while `product` is now required on `productRelease`. `tea-product.md` and `tea-product-release.md` were aligned; the README was not.
> 
> **2. `collection.examples[0]` — both artifacts fail `artifact.required`** (`[uuid, type, formats, createdDate, version]`):
> 
> * `artifacts[0]` has only `formats, name, type, uuid` — missing `createdDate` and `version`
> * `artifacts[1]` has `formats, name, type, uuid, version` — missing `createdDate`
> 
> **3. `component-release-with-collection.examples[0]` — three failures:**
> 
> * the embedded `release` has `createdDate, identifiers, releaseDate, uuid, version` — missing required `component`. This is the Tomcat **11.0.7** example; `7bdfff0` added `component` to the pre-release `11.0.0-M26` example but not to this one.
> * `latestCollection.artifacts[0]` and `latestCollection.artifacts[1]` — both missing `createdDate` and `version`

These should be fixed now.

oej:
I will go ahead and merge now. Any new issues will have to be handled in new issues/PRs. Thank you for your reviews!

PR #265, opened 2026-09-10
feat: discovery by PURL as an alternative to TEI

taleodor:
/discovery accepts either `tei` or `purl`; exactly one shall be given, a request with neither or both is rejected with 400. OpenAPI cannot express that constraint across query parameters, so `tei` becomes optional and the rule is stated in the descriptions.

Discovery by TEI resolves across TEA servers through the domain the TEI carries; discovery by PURL resolves within the inventory of the server receiving the request, since a PURL carries no domain.

Spec only; documentation unchanged to prevent conflicts with #261 . Validated with openapi-generator v7.12.0 (validate; Go batch generation as CI runs it).

### DIFF
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -763,20 +763,20 @@ paths:
 
         Discovery by TEI resolves the product across TEA servers, using the domain carried
         by the TEI. Discovery by PURL resolves the product within the inventory of the TEA
-        server that receives the request; a PURL carries no domain.
+        server that receives the request.
       operationId: discoveryByTei
       parameters:
         - name: tei
           in: query
           required: false
-          description: Transparency Exchange Identifier (TEI) for the product being discovered. Provide the TEI as a URL-encoded string per RFC 3986, RFC 3987. Mutually exclusive with `purl`.
+          description: Transparency Exchange Identifier (TEI) for the product being discovered. Provide the TEI as a URL-encoded string per RFC 3986, RFC 3987 or the PURL under purl parameter.
           schema:
             type: string
             example: urn%3Atei%3Auuid%3Aproducts.example.com%3Ad4d9f54a-abcf-11ee-ac79-1a52914d44b
         - name: purl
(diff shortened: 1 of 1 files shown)


spec/openapi.yaml
```diff
@@ -754,16 +754,32 @@ paths:
...
+      description: |
+        Discovery endpoint which resolves an identifier into product release UUID.
+
+        Exactly one of the `tei` and `purl` query parameters shall be provided. A request
+        with neither, or with both, is rejected with `400`. OpenAPI cannot express this
+        constraint on query parameters, so both are declared optional here.
+
+        Discovery by TEI resolves the product across TEA servers, using the domain carried
+        by the TEI. Discovery by PURL resolves the product within the inventory of the TEA
+        server that receives the request; a PURL carries no domain.
```
oej: We can't assure that no PURL carries a domain. The new one being discussed will carry a domain.

spec/openapi.yaml
```diff
@@ -754,16 +754,32 @@ paths:
...
       operationId: discoveryByTei
       parameters:
         - name: tei
           in: query
-          required: true
-          description: Transparency Exchange Identifier (TEI) for the product being discovered. Provide the TEI as a URL-encoded string per RFC 3986, RFC 3987.
+          required: false
+          description: Transparency Exchange Identifier (TEI) for the product being discovered. Provide the TEI as a URL-encoded string per RFC 3986, RFC 3987. Mutually exclusive with `purl`.
```
oej: I would say "Provide the TEI or the PURL as an URL-encoded string"... and remove the last sentence.

oej (changes requested):

spec/openapi.yaml
taleodor: good point, removed that note

spec/openapi.yaml
taleodor: updated

PR #270, opened 2026-09-11
Align OpenAPI security with open and mixed TEA servers

### LINKED ISSUE #269: Align OpenAPI security with open and mixed TEA servers
## Summary

`auth/readme.md` describes open and mixed servers, but the OpenAPI document currently declares global `security: Bearer` only. The reusable `401` response is also unused on resource operations, and there is no `403`.

## Suggestion

1. Global security as Bearer **or** empty (`{}`), so anonymous calls are representable.
2. Declare `401` / `403` on resource operations (runtime still decides which endpoints actually require a token).
3. Clarify that `/token`’s empty security requirement is for alternate client authentication (for example, mutual TLS, `private_key_jwt`, or credentials in the request body), **not** anonymous `client_credentials`.
4. Note that servers may conceal unauthorized resources with `404`.

Mehrn0ush:
## Summary
- Represent open/mixed servers in OpenAPI with Bearer or anonymous (`{}`) security
- Declare `401` / `403` on resource operations, and clarify concealing `404` for unauthorized access
- Document that `/token`’s empty security requirement covers alternate client auth (mutual TLS, `private_key_jwt`, or credentials in the request body), not anonymous `client_credentials`

Thanks to @taleodor for the review feedback.

Closes #269

### DIFF
--- a/auth/readme.md
+++ b/auth/readme.md
@@ -203,7 +203,10 @@ certificate. This restates RFC 6749 section 3.2 and RFC 6750 section 5.
 ## Client flow
 
 Whether a server requires authentication is discovered by using it, not by configuration and not by
-probing the token endpoint. The complete flow for a client that does not know in advance:
+probing the token endpoint. In the OpenAPI document, resource operations list both Bearer
+authentication and an empty security requirement so open and mixed deployments are representable;
+a `401` with a `WWW-Authenticate: Bearer` challenge on a given endpoint is still the runtime signal
+that a token is required there. The complete flow for a client that does not know in advance:
 
 1. The client sends the resource request it wants, with no credentials.
 2. If the server does not require authentication for that endpoint, it answers `200` with the
@@ -212,7 +215,9 @@ probing the token endpoint. The complete flow for a client that does not know in
    the `Bearer` scheme, as described in RFC 6750 section 3. This challenge is the only signal a
    client needs.
 4. The client calls `POST /token` with its credential - for the baseline, the API key over HTTP
-   Basic - and receives an access token.
+   Basic - and receives an access token. Unauthenticated `client_credentials` requests are not
+   permitted; an empty OpenAPI security requirement on `/token` allows alternate client
+   authentication (for example mutual TLS), not anonymous token issuance.
 5. The client repeats the resource request with `Authorization: Bearer <access_token>`, and presents
    the same token on subsequent requests until it expires or is rejected.
 6. When a later request fails with `401` and `error="invalid_token"`, the client obtains a fresh
@@ -261,7 +266,8 @@ authentication:
 
 Authorization - which of the protected data an authenticated client may see - is the server's
 decision and is not constrained by this specification; a client with a valid token may still receive
-a filtered view, or `403`/`404` for individual objects.
+a filtered view, `403 Forbidden`, or a concealing `404 Not Found` for individual objects. Clients
+shall not infer from `404` alone whether the object is absent or withheld.
 
 ### Protected resource metadata
 
--- a/auth/readme.md
+++ b/auth/readme.md
@@ -217,7 +217,8 @@ that a token is required there. The complete flow for a client that does not kno
 4. The client calls `POST /token` with its credential - for the baseline, the API key over HTTP
    Basic - and receives an access token. Unauthenticated `client_credentials` requests are not
    permitted; an empty OpenAPI security requirement on `/token` allows alternate client
-   authentication (for example mutual TLS), not anonymous token issuance.
+   authentication (for example, mutual TLS, `private_key_jwt`, or credentials in the
+   request body), not anonymous token issuance.
 5. The client repeats the resource request with `Authorization: Bearer <access_token>`, and presents
    the same token on subsequent requests until it expires or is rejected.
 6. When a later request fails with `401` and `error="invalid_token"`, the client obtains a fresh
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -840,9 +840,10 @@ paths:
 
         The empty security requirement on this operation does not authorize anonymous
         `client_credentials` issuance. It only allows the alternate client-authentication
-        methods described above (for example mutual TLS). For the baseline
-        `client_credentials` grant, servers shall require HTTP Basic client authentication
-        and shall reject unauthenticated token requests with `401`.
+        methods described above (for example, mutual TLS, `private_key_jwt`, or
+        credentials in the request body). For the baseline `client_credentials` grant,
+        servers shall require HTTP Basic client authentication and shall reject
+        unauthenticated token requests with `401`.
       operationId: requestToken
       security:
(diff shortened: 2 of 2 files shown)


taleodor (approved):

auth/readme.md
```diff
@@ -203,7 +203,10 @@ certificate. This restates RFC 6749 section 3.2 and RFC 6750 section 5.
 ## Client flow
 
 Whether a server requires authentication is discovered by using it, not by configuration and not by
-probing the token endpoint. The complete flow for a client that does not know in advance:
+probing the token endpoint. In the OpenAPI document, resource operations list both Bearer
```
oej: I think we don't have to refer to the OpenAPI document, but instead say "In TEA"...

spec/openapi.yaml
```diff
@@ -2023,6 +2138,10 @@ components:
         otherwise invalid. Servers shall include a `WWW-Authenticate` header as defined
         in RFC 6750 section 3. On the `invalid_token` error a client may obtain a fresh
         token from `/token` and retry the request once (RFC 6750 section 3.1).
+
+        Open servers that require no authentication on any endpoint shall not return this
+        status. On a mixed server, only protected endpoints return `401`; open endpoints
```
oej: I am not happy with the term "mixed server". Maybe write

"On a TEA server where some data is available without authorisation, but not all, protected endpoints return"...

spec/openapi.yaml
```diff
@@ -2458,6 +2594,15 @@ components:
         A TEA access token obtained from `/token`, presented as
         `Authorization: Bearer <access_token>` (RFC 6750). This is the only credential
         accepted on TEA endpoints other than `/token`.
+
+        Resource operations declare both this requirement and an empty alternative so
+        that open and mixed servers are representable: a client may call without a
```
oej: "open and mixed servers"... maybe

"so that a server can select which objects that require authentication"

spec/openapi.yaml
```diff
@@ -2458,6 +2594,15 @@ components:
...
+        token. Whether a given endpoint actually requires a token is discovered at
+        runtime from a `401` Bearer challenge (see `401-unauthorized`), not from this
+        document alone. Open servers shall ignore a presented Bearer token; mixed
```
oej: Remove "mixed"

oej:
In general I don't like the term "open" and "mixed" - it's not normative language in an API spec. It's a server that doesn't require auth and thus have no authz. And a server that wants to protect some objects, but not all, with both auth and authz. Please rewrite

auth/readme.md
Mehrn0ush: Agreed

spec/openapi.yaml
Mehrn0ush: Updated

spec/openapi.yaml
Mehrn0ush: Replaced

spec/openapi.yaml
Mehrn0ush: Removed. That paragraph now describes a server with no authentication vs one that protects some objects but not all.

Mehrn0ush:
Rewrote throughout auth/readme.md and the OpenAPI security/401 prose. Terminology is now “server that requires no authentication” and “server that protects some objects,” matching auth vs authz as you described. Also renamed the former “Mixed servers” section accordingly.

taleodor:
Thanks for the rewrite! The global `security: [{bearerAuth: []}, {}]` is the right OpenAPI idiom for optional auth, and I checked the head branch: all 27 resource operations now declare `401` + `403`, `/token` keeps `[basicAuth, {}]`. Structurally clean. A few things I'd like resolved before merge:

### 1. Concealing `404` for *unauthenticated* clients contradicts the client flow

`auth/readme.md` step 3 says the `401` Bearer challenge "is the only signal a client needs", and "Servers that protect some objects" says protected endpoints "answer `401` … when no valid token is presented, which is how a client learns, per endpoint, that a token is needed."

The new `404-object-by-id-not-found` description allows a concealing `404` "for example when the client is **unauthenticated** or authenticated but not authorized". Under the flow this spec defines (step 1: send with no credentials), an entitled client requesting a concealed object gets `404`, concludes the object does not exist, and never authenticates. The concealment is indistinguishable from absence by design, so the flow fails silently.

This also interacts with #296: a TEA `404` on `/discovery` now means "identifier unknown — stop, do not fail over". An unauthenticated concealment `404` there makes a client abandon a TEI it is entitled to.

**Suggestion:** limit concealment to the authenticated-but-unauthorized case (`403` vs concealing `404` is a legitimate server choice there), and keep "no valid token on a protected object ⇒ `401`" as a `shall`. Concretely, drop "unauthenticated" from the `404-object-by-id-not-found` description and from the corresponding readme sentence.

### 2. `403-forbidden` body schema

`403-forbidden` uses `error-response`, whose `error` enum is only `OBJECT_UNKNOWN | OBJECT_NOT_SHAREABLE`. A `403` carrying `OBJECT_UNKNOWN` is contradictory. If `OBJECT_NOT_SHAREABLE` is the intended value, please state it and add an `example`. Also, `401-unauthorized` has `application/json: {}` (no schema) — the two should be consistent: either both carry `error-response` or neither.

### 3. Terminology

`401-unauthorized`: "where some data is available without **authorization**, but not all" → should be *authentication*, per the auth/authz distinction the readme was just rewritten around.

### 4. Smaller things

- `/token`: "shall reject unauthenticated token requests with `401`" — RFC 6749 §5.2 defaults to `400` `invalid_client`; `401` is MUST only when the client attempted an `Authorization` header, otherwise MAY. Requiring `401` is a valid tightening given Basic is mandatory, but it should name `invalid_client` and the `WWW-Authenticate: Basic` challenge — worth confirming `401-token-error` already does.
- `private_key_jwt` is OpenID Connect vocabulary; the rest of the document cites RFCs — "JWT client assertion (RFC 7523)" would fit better.
- Verbal forms (ISO/Ecma style): "Unauthenticated `client_credentials` requests are not permitted" → "shall not be accepted".

Mehrn0ush:
Thanks @taleodor 

1. Concealing 404 vs client flow
Agreed. Concealment is limited to authenticated-but-unauthorized. Protected objects with no valid token shall return 401, not a concealing 404. Dropped “unauthenticated” from the 404 description and aligned the auth readme / bearerAuth text.

2. 403 body schema
Made 403-forbidden consistent with 401-unauthorized: both use application/json: {} (no error-response). A 403 carrying OBJECT_UNKNOWN was the wrong model.

3. Terminology
Updated to “without authentication” in 401-unauthorized.

4. Smaller things
/token: baseline Basic failures → 401 + invalid_client + WWW-Authenticate: Basic (via 401-token-error). Replaced private_key_jwt with “JWT client assertion (RFC 7523)”. “shall not be accepted” for unauthenticated client_credentials.

taleodor:
LGTM now.

PR #272, opened 2026-09-11
Align /token responses with RFC 6749 cache and Basic encoding

### LINKED ISSUE #271: Align /token with RFC 6749 Pragma and Basic credential encoding
## Summary

Two small RFC 6749 gaps in the current `/token` contract:

1. **Token response caching** — OpenAPI declares `Cache-Control: no-store` on successful token responses, but not `Pragma: no-cache`. [RFC 6749 section 5.1](https://www.rfc-editor.org/rfc/rfc6749.html#section-5.1) requires both for responses that include tokens (or other sensitive values).

2. **Basic client authentication** — The `basicAuth` scheme describes API key id/secret as username/password, but does not state that each value is `application/x-www-form-urlencoded` encoded before `userid:password` and Base64, as required by [RFC 6749 section 2.3.1](https://www.rfc-editor.org/rfc/rfc6749.html#section-2.3.1). That encoding matters for secrets containing `:`, `@`, or other reserved characters. (`auth/readme.md` already notes this; OpenAPI should match.)

## Proposed fix

- Add `Pragma: no-cache` beside `Cache-Control: no-store` on `token-issued` (and align the auth doc example).
- Document the §2.3.1 form-encoding step in the `basicAuth` security scheme description.

Mehrn0ush:
## Summary
- Add `Pragma: no-cache` on successful token responses alongside existing `Cache-Control: no-store`, per RFC 6749 section 5.1
- Document in the `basicAuth` scheme that API key identifier and secret are each `application/x-www-form-urlencoded` encoded before `userid:password` and Base64, per RFC 6749 section 2.3.1
- Align the token success example in `auth/readme.md` with the same `Pragma` header

Closes #271

## Context
RFC 6749 section 5.1 requires both `Cache-Control: no-store` and `Pragma: no-cache` on responses that include tokens. OpenAPI previously only declared `Cache-Control`.

RFC 6749 section 2.3.1 requires form-encoding each credential before building the Basic header. That step matters for secrets containing `:`, `@`, or other reserved characters. `auth/readme.md` already described it; the OpenAPI `basicAuth` description did not.

### DIFF
--- a/auth/readme.md
+++ b/auth/readme.md
@@ -101,6 +101,7 @@ A successful response is the standard OAuth 2.0 token response (RFC 6749 section
 HTTP/1.1 200 OK
 Content-Type: application/json;charset=UTF-8
 Cache-Control: no-store
+Pragma: no-cache
 
 {
   "access_token": "2YotnFZFEjr1zCsicMWpAA",
--- a/spec/openapi.yaml
+++ b/spec/openapi.yaml
@@ -2149,6 +2149,13 @@ components:
           schema:
             type: string
             example: no-store
+        Pragma:
+          description: |
+            Servers shall set `no-cache` on token responses (RFC 6749 section 5.1), for
+            compatibility with HTTP/1.0 caches.
+          schema:
+            type: string
+            example: no-cache
       content:
         application/json:
           schema:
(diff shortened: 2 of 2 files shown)


oej:
@taleodor what do you think?

taleodor:
I think we shouldn't merge this. Pragma is deprecated in RFC 9111 and dropped in OAuth 2.1. I don't think we should mandate a legacy header.

taleodor:
Alternatively, maybe remove Pragma completely from PR and just keep encoding part?

Mehrn0ush:
Good catch — agreed. RFC 6749 includes Pragma: no-cache for HTTP/1.0 compatibility, but RFC 9111 deprecates it and OAuth 2.1 dropped it in favor of Cache-Control: no-store alone. I’ll remove the Pragma requirement from this PR and keep the Basic credential encoding clarification.

Mehrn0ush:
Thanks  @taleodor — that makes sense. I’ve removed the Pragma requirement from this PR and kept the Basic form-urlencoded encoding clarification.

oej:
Are we ready to merge @taleodor ?

taleodor (approved):

PR #281, opened 2026-09-14
Docs/align tea api version grammar

### LINKED ISSUE #280: Align TEA API version grammar and /v{version} URL construction
## Summary

Advertised TEA API versions claim SemVer 2.0.0, but the `.well-known` schema pattern, examples, and OpenAPI `tea-server-info.versions` do not agree — so clients cannot implement “highest mutual version + `/v{version}`” consistently.

## Current gaps

- `discovery/tea-well-known.schema.json` accepts non-SemVer values such as `1.0`, `01.0.0`, and invalid prereleases, and rejects valid SemVer build metadata.
- The schema example includes `"1.0"` while describing SemVer 2.0.
- OpenAPI `tea-server-info.versions` items are unconstrained strings.
- The OpenAPI `servers` example uses `/v1`, which does not match discovery’s “append `/v` + exact advertised version” rule (`/v1.0.0`, `/v0.2.0-beta.2`, …).

## Proposed direction

- Use one SemVer 2.0.0 grammar in the well-known schema and OpenAPI: `MAJOR.MINOR.PATCH` with optional prerelease.
- Do not advertise build metadata (`+…`) in API path versions (awkward in URL paths).
- Fix examples to full versions (e.g. `1.0.0`).
- State clearly: select the highest mutually supported version; append `/v` followed by that exact advertised string.
- Align the local OpenAPI server example with `info.version` (e.g. `/v0.5.0`).

Does this direction look right before a PR ?

Mehrn0ush:
## Summary
- Require full SemVer `MAJOR.MINOR.PATCH` for advertised API versions in the well-known schema and OpenAPI
- Switch version examples to `1.0.0` (including `/v1.0.0` construction)
- Keep URL rule: highest mutual version, then append `/v` + that exact string

Closes #280.

Thanks @taleodor


discovery/readme.md
```diff
@@ -287,22 +285,23 @@ in the TEI URI running on the default port to enable discovery of the API server
 ## Connecting to the API
 
 Clients must pick any one of the endpoints listed in the `.well-known/tea` json
-response. The client MUST pick an endpoint with the at least one version that is
-supported by the client is using. The client MUST prioritize endpoints with the
+response. The client MUST pick an endpoint with at least one version that is
```
taleodor: MUST -> SHALL

discovery/readme.md
```diff
@@ -287,22 +285,23 @@ in the TEI URI running on the default port to enable discovery of the API server
...
+supported by the client. The client MUST prioritize endpoints with the
```
taleodor: Same, MUST -> SHALL

discovery/readme.md
```diff
@@ -287,22 +285,23 @@ in the TEI URI running on the default port to enable discovery of the API server
...
-between 0 and 1).
-
-The client must then construct the full URL to the API by appending the
-"/v" plus one of the versions listed in the `versions` array of the selected endpoint,
-plus "/discovery?tei=", plus the TEI that is url-encoded according to [RFC3986]
+Advertised API versions are full SemVer `MAJOR.MINOR.PATCH`. If there are several
+endpoints like these and if the priority field is present, the client SHOULD pick
+the endpoint with the highest priority value (a float between 0 and 1).
+
+The client MUST then construct the full URL to the API by selecting that highest
```
taleodor: Same.

discovery/readme.md
```diff
@@ -287,22 +285,23 @@ in the TEI URI running on the default port to enable discovery of the API server
...
+mutually supported version and appending `/v` followed by that exact advertised
+version string (for example `/v1.0.0`), then
+`/discovery?tei=` plus the TEI that is url-encoded according to [RFC3986]
 and [RFC3986]).
 
 Examples:
 1. For TEI `urn:tei:uuid:products.example.com:d4d9f54a-abcf-11ee-ac79-1a52914d44b`
-`https://api.teaexample.com/v0.2.0-beta.2/discovery?tei=urn%3Atei%3Auuid%3Aproducts.example.com%3Ad4d9f54a-abcf-11ee-ac79-1a52914d44b`
+`https://api.teaexample.com/v1.0.0/discovery?tei=urn%3Atei%3Auuid%3Aproducts.example.com%3Ad4d9f54a-abcf-11ee-ac79-1a52914d44b`
```
taleodor, line 305: TEI is now URL

discovery/tea-well-known.schema.json
```diff
@@ -32,17 +32,15 @@
...
             "type": "string",
-            "pattern": "^\\d+\\.\\d+(?:\\.\\d+)?(?:-[0-9A-Za-z.-]+)?$",
+            "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)$",
             "examples": [
-              "0.1.0-beta.1",
-              "0.2.0-beta.2",
               "1.0.0"
             ],
-            "description": "TEA OpenAPI Spec Version identifier, conforms to SemVer 2.0 (https://semver.org/)."
+            "description": "TEA OpenAPI Spec version identifier as full SemVer 2.0.0 MAJOR.MINOR.PATCH (https://semver.org/)."
```
taleodor: You're changing regex to digits only but prose says "full SemVer". Full SemVer may have modifier and metadata. I suggest reverting this back actually and keeping previous regex (that was representing full SemVer per spec).

spec/openapi.yaml
```diff
@@ -1609,12 +1609,17 @@ components:
           format: uri
           example: https://api.teaexample.com
         versions:
-          description: Supported TEA API versions at this server without v prefix
+          description: |
+            Supported TEA API versions at this server, without a leading `v`. Each value
+            is full SemVer 2.0.0 `MAJOR.MINOR.PATCH`. Clients select the highest mutually
+            supported version and append `/v` followed by that exact string.
```
taleodor: See previous comment re: SemVer.

spec/openapi.yaml
```diff
@@ -1609,12 +1609,17 @@ components:
...
           type: array
           minItems: 1
           items:
             type: string
-          example: ["0.2.0-beta.2", "1.0.0"]
+            pattern: '^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)$'
+            description: TEA OpenAPI Spec version identifier (SemVer 2.0.0 MAJOR.MINOR.PATCH)
+          example: ["1.0.0"]
```
taleodor, line 1623: See previous comment re: SemVer.

taleodor (changes requested):

discovery/tea-well-known.schema.json
Mehrn0ush: agreed.I’ll restore a SemVer-accurate pattern and keep the 1.0.0 examples

discovery/readme.md
Mehrn0ush: This branch is still based on current main (URN). Would you prefer I update the TEI examples in this PR to the URL form now, or wait until #261 merges and then rebase/apply that change here?

Mehrn0ush:
Thanks @taleodor  — pushed a follow-up for the vocabulary and semVer; 
On TEI-as-URL: as this branch is still on current main (URN). Happy to either wait for #261 to merge and then rebase, or update the TEI examples here now if you’d rather not wait — which do you recommend?

discovery/readme.md
oej: Wait and rebase after merge of URN removal

Mehrn0ush:
rebased.

taleodor:
still see conflicts

Mehrn0ush:
fixed

taleodor:
Thanks! Rebase is clean and the SemVer regex checks out (I tested it: rejects `1.0`, `01.0.0`, `v1.0.0`; accepts prerelease and build metadata; both `.well-known` example endpoints validate). Three remaining items:

### 1. Build-metadata tie is unresolved

> Versions that differ only in build metadata have equal SemVer precedence; when choosing among them, the client shall preserve the exact advertised spelling of the selected string.

This says how to *spell* the choice but not how to *make* it — if an endpoint advertises `1.0.0+a` and `1.0.0+b`, "highest mutually supported" is a tie and the text is silent. Since SemVer 2.0.0 §10 makes these the same version, an arbitrary choice is the consistent answer. The mutual-support check also needs to ignore build metadata, otherwise a client that "supports 1.0.0" would string-compare and treat `1.0.0+b` as unsupported. Suggested replacement:

> Build metadata is ignored when comparing versions (SemVer 2.0.0 §10), both for precedence and for determining whether the client supports a version. If the highest mutually supported version is advertised with more than one build-metadata spelling, the client may select any of them, and shall use the selected string exactly as advertised when constructing the path (for example `/v1.0.0+build.5`; `+` is a valid path character per RFC 3986 and shall not be percent-encoded).

### 2. Pre-existing typo in the rewritten paragraph

> …plus the TEI that is url-encoded according to [RFC3986]
> and [RFC3986]).

Duplicate reference and a stray `)`. Since this paragraph is being rewritten anyway, please fix it here: "…the TEI, url-encoded according to [RFC3986]."

### 3. Verbal-form casing

The new text uses lowercase `shall` (correct for ISO/Ecma style), but the same paragraph still carries "the client **SHOULD** pick the endpoint with the highest `priority`". Pre-existing, but now the only uppercase form in a rewritten paragraph — please lowercase it.

Mehrn0ush:
Thanks — addressed all three: build-metadata ignored for support/precedence with arbitrary tie-break + exact path spelling (including unencoded +);
fixed the duplicate RFC3986 typo;
lowercased should for priority.

taleodor (approved):

PR #284, opened 2026-09-14
Use logical values for TEI, PURL, and mediaType query params

### LINKED ISSUE #283: Avoid double-encoding of TEI, PURL, and mediaType query parameters
## Summary

Some OpenAPI query-parameter examples for tei, purl, and artifact mediaType appear to contain values that have already been percent-encoded for URI transport.

## Current issues

- `/discovery` `tei` / `purl` examples use forms like `urn%3Atei%3A...` and `pkg%3Amaven%2F...`
- Artifact / signature `mediaType` examples use `application%2Fvnd.cyclonedx%2Bjson`
- Descriptions tell callers to “provide a URL-encoded string,” which conflates the parameter value with HTTP serialization

## Proposed direction

- Use raw TEI, PURL, and media-type values in parameter schemas/examples
- State that clients apply query-parameter serialization once, and servers interpret after one query-decoding step (preserving escapes that are part of a PURL value)
- Show percent-encoded forms only as example request lines after serialization

Does this direction look right?

Mehrn0ush:
## Summary
- Use logical (non-pre-encoded) examples for `/discovery` `tei`/`purl` and artifact `mediaType` query parameters
- Clarify: encode once on the client; decode once on the server
- Keep percent-encoded forms only in “after serialization” request-line examples
Closes #283.


spec/openapi.yaml
```diff
@@ -772,22 +772,38 @@ paths:
...
           required: false
-          description: Transparency Exchange Identifier (TEI) for the product being discovered. Provide the TEI as a URL-encoded string per RFC 3986, RFC 3987 or the PURL under purl parameter.
+          description: |
+            Transparency Exchange Identifier (TEI) for the product being discovered, as the
+            logical TEI string. Do not pre-encode for the schema example or client model;
+            apply query-parameter encoding once when forming the HTTP request.
           schema:
             type: string
-            example: urn%3Atei%3Auuid%3Aproducts.example.com%3Ad4d9f54a-abcf-11ee-ac79-1a52914d44b
+            example: urn:tei:uuid:products.example.com:d4d9f54a-abcf-11ee-ac79-1a52914d44b
```
taleodor, line 795: this needs to use new URI syntax

taleodor (changes requested):
Generally agree with direction, but this should be rebased after #261 merges.

Mehrn0ush:
Thanks @taleodor — agreed. I’ll hold this until #261 merges, then rebase and switch the tei example to the new URI syntax.

oej (approved):
Looks good to me. Normative changes, so urgency

taleodor:
See below - generally, 2 and 3 should be addressed, then we should wait on #261 merge before merging this.

**1. This collides with #261 on the same line, and carries the URN form.** This PR sets:

```yaml
example: urn:tei:uuid:products.example.com:d4d9f54a-abcf-11ee-ac79-1a52914d44b
```

while #261 sets that same line to `tei://products.example.com/uuid/47f51950-…`. It also adds a new URN reference in the serialization block at line 782, which #261 will make stale. Merge order matters: land #261 first, then rebase this and switch both the example and the serialization line to the `tei://` form. If this lands first, whoever resolves the conflict has to remember not to keep the URN form.

**2. The truncated UUID is being propagated.** `d4d9f54a-abcf-11ee-ac79-1a52914d44b` has 11 hex digits in the final group rather than 12. This PR now carries it in two places: the `example` value and line 782. It is the same defect flagged on #261 lines 333-334, so it is worth fixing here rather than copying it forward — #261's replacement UUID is already correct.

**3. Keyword inconsistency within the PR.** The `tei` and `purl` parameters use singular `example:` under `schema:`, while the two `mediaType` parameters use the array form:

```yaml
schema:
  type: string
  examples:
    - application/vnd.cyclonedx+json
```

The file declares `jsonSchemaDialect: …/2020-12`, where a Schema Object's keyword is `examples` (an array); `example` is the legacy OpenAPI spelling. Since both halves of this PR touch the same concern, harmonising on `examples:` would be a two-line change, and it matches what #264 does across the component schemas.

Mehrn0ush:
Thanks,
Addressed (2) and (3) in a follow-up commit:
 full UUID (`…44b1`) in both places, and `examples:` for `tei`/`purl`. Still holding the `tei://` switch and merge until #261 lands, then will rebase.

oej:
#261 is now merged, so let's focus on fixing and merging this

Mehrn0ush:
Rebased onto main after #261. TEI examples now use `tei://` (logical + serialized); kept `examples:` and one-shot encoding guidance.

PR #293, opened 2026-09-16
Define checksum hex encoding and fix invalid examples

### LINKED ISSUE #290: Define checksum encoding and fix invalid checksum examples
## Summary

Checksums appear in published OpenAPI tables, but `algValue` encoding and hashed-bytes scope are underspecified, and several examples contradict the `checksum-type` enum. Invalid examples poison Ecma/table generation the same way as schema mismatches.

## Gaps

1. Encoding / hashed bytes — `algValue` is a free `string` (`description: Checksum value`) with no `format`, `pattern`, or length constraint. Prose does not say whether values are lowercase hex (vs Base64), length expectations per algorithm, or which bytes are hashed (artifact/distribution payload vs other).

2. Invalid examples — Enum is hyphenated (`SHA-256`, `SHA-512`, …), but multiple distribution examples still use `SHA_256` / `SHA_512`. Related Tomcat examples also mix `11.0.7` release metadata with `11.0.6` distribution PURLs/URLs (e.g. release `version: "11.0.7"` with nested `tomcat@11.0.6?type=zip` and a URL path containing `/11.0.7/tomcat-11.0.6.zip`).

## Proposed normative direction

- `algValue` **shall** be lowercase hexadecimal for the digest of the identified bytes; document length (or pattern) per `algType` where practical.
- State which bytes are covered for distribution vs artifact checksums.
- Examples **shall** use enum-valid `algType` values and internally consistent version/PURL/URL data.

OK to ship encoding and example fixes in one OpenAPI PR, or split “normative encoding” from “example cleanup”?

Mehrn0ush:
## Summary
- Document `algValue` as lowercase hex and which bytes distribution/artifact checksums cover
- Fix invalid `SHA_*` examples and Tomcat 11.0.7/11.0.6 mismatches
Closes #290.


spec/openapi.yaml
```diff
@@ -1566,18 +1568,34 @@ components:
...
+            `mediaType`), not the detached signature.
           items:
             "$ref": "#/components/schemas/checksum"
     checksum:
       type: object
+      description: |
+        Integrity digest for identified content bytes. `algValue` is the lowercase
+        hexadecimal encoding of the digest produced by `algType` over those bytes.
+        The hex length shall match the algorithm's digest size (for example 32 for
+        MD5, 40 for SHA-1, 64 for SHA-256 / SHA3-256 / BLAKE2b-256 / BLAKE3,
```
oej: MD5 is not supported

oej:
One small thing to fix

spec/openapi.yaml
Mehrn0ush: Thanks. 
fixed.

oej (approved):

PR #294, opened 2026-09-16
fix: bound collection/artifact revisions and define latest (#287)

### LINKED ISSUE #287: Constrain collection/artifact revision bounds and define "latest"
## Summary

Collection and artifact revision numbers are described as starting at 1 and supporting successive immutable revisions, but OpenAPI path parameters and schema integers still admit `0` and negatives, and “latest” is not defined as a selection rule. #264 requiring `version` does not constrain its range.

## Current gaps

- Path parameters `collectionVersion` and `artifactVersion` are plain `integer` with no `minimum: 1` (e.g. collection and artifact download/metadata paths).
- `collection.version` prose says versions start with 1, but the schema has no lower bound.
- `artifact.version` has `default: 1` and talks about successive immutable revisions, but does not require positive values or define how “latest” is chosen.
- Six versioned paths plus the collection/artifact models should share one positive-integer revision rule so clients and servers agree.

## Proposed direction

- Introduce a shared positive-integer revision schema (or equivalent `minimum: 1`) for all versioned path params and for `collection.version` / `artifact.version`.
- State normatively:
  - Revision numbers SHALL start at 1 and increase monotonically.
  - **Latest** means the greatest published revision number.
  - Published content/signatures for a given revision SHALL NOT change.
- Clarify which metadata (e.g. expiring external URLs) MAY be refreshed without bumping the content revision.

## Relation to other work

- **#264 / #259:** requiring `version` is complementary; it does not close bounds or latest-selection. This issue can stay separate so #264 remains focused.
- Related download/`/latest` endpoints already assume a latest revision exists; this issue defines how that revision is identified.

## Out of scope

- Whether a release may exist without a collection (#283-style / release-without-collection discussion)
- Checksum encoding / invalid examples (separate findings)
- Pagination filter semantics

Does this direction look right ?

Mehrn0ush:
## Summary
- Add shared `content-revision` schema (`minimum: 1`) and wire it into collection/artifact path params and model `version` fields
- Define `latest` endpoints as the greatest published revision
- Clarify that published content/signatures for a revision shall not change; non-content metadata may refresh without bumping revision
- 
Closes #287


spec/openapi.yaml
```diff
@@ -448,9 +456,10 @@ paths:
         - name: collectionVersion
           in: path
           required: true
-          description: Version of TEA Collection
+          description: |
+            Revision of the TEA Collection. Shall be an integer greater than or equal to 1.
```
oej, line 460: "Should be incremented by 1 for every new collectionVersion."

spec/openapi.yaml
```diff
@@ -951,6 +970,19 @@ components:
       description: A UUID
       format: uuid
       pattern: "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+    content-revision:
```
oej, line 978: If we add this, do we have API support for getting latest revision and previous revisions?

Why not call it "version" like other objects?

spec/openapi.yaml
Mehrn0ush: Good catch — I’ll state that each new published collection revision increments by 1 (starting at 1), not only that the value is ≥ 1.

spec/openapi.yaml
Mehrn0ush: Latest and prior revisions are already reachable: …/collection/latest and …/collection/{collectionVersion}, and the same pattern for artifacts (/artifact/{uuid}/latest vs /artifact/{uuid}/{artifactVersion}). This change only constrains the integer and defines “latest” as the greatest published revision.

The object property stays version. content-revision is only the shared schema name so we don’t overload SemVer string version used elsewhere.  Happy to rename the schema if you prefer something clearer, e.g. content-version, or any other name you think fits better.

oej:
@taleodor - what do you think? Ready to merge?

taleodor:
I would close this PR, I don't think we need this:

1. Adding `minimum: 1` doesn't buy us much, non-existing revision would just yield 404
2. Regarding `latest`, we already say (on main) that versions start 1 and increment. There is pretty clear wording.
3. Immutability is actually explained in multiple places (including etag, cache-control, signature endpoint)

The good part is sentence that pre-signed external URL may change without bumping the revision. But I suggest making this a separate PR.

Wording suggestions (for that separate PR):

1. 
old
```
            the URL is pre-signed it may expire; clients should not retain it beyond the
            freshness lifetime of the response that carried it.
```

new
```
            the URL is pre-signed it may expire; clients should not retain it beyond the
            freshness lifetime of the response that carried it. A server may replace an
            expired or renewed URL in place: that is not a new artifact revision, and the
            content and checksums it points to are unchanged.
```

2. 
old

```
 description: Get metadata for latest revision of a specific TEA Artifact
```

new
```
      description: |
        Get metadata for the latest revision of a specific TEA Artifact, that is, the
        revision with the greatest published `version`.
 ```
 
 3. 
 old
  ```
       description: Get the latest TEA Collection belonging to the TEA Component Release
 ```
 new
 ```
       description: |
        Get the latest TEA Collection belonging to the TEA Component Release, that is,
        the collection with the greatest published `version`.
 ```

Mehrn0ush:
Thanks @taleodor — fair on not wanting a large restatement. agreed.

The reason I opened #287 wasn’t only minimum: 1. On main today, path params are still unconstrained integers, collection /latest doesn’t define the selection rule, and artifact latest download is circular (“whichever /latest reports”). That said, I agree the shared schema / immutability restatement may not be worth the weight if invalid revisions already 404 and immutability is covered elsewhere.

taleodor:
What's the issue with unconstrained integers? I mean what kind of problem it addresses beyond client getting 404 (but they still may get it if they put some arbitrarily large integer for a version that doesn't exist)

For latest - the server just sends the latest revision via simple integer comparison. If this is not clear, I think a sentence in docs would be appropriate, but not large rewrite. But to me this reads fairly obvious. We have single component version which always increments, so we just pick largest - do you see what this can be confused with?

RE: and artifact latest download is circular (“whichever /latest reports”) -> I don't mind clarification wording change there, I think it's fine, I want to avoid schema rewrite.

Mehrn0ush:
Thanks — yes, you’re right. Unconstrained integers don’t really buy us anything if a bad or missing revision still just 404s. And “latest = largest version” is pretty obvious once versions only go up; I don’t think people would read it as something else.

The only bit I’d still tidy is the circular “whichever /latest reports” wording on artifact download, plus your pre-signed URL note.

 I’ll close this PR and do a small wording-only follow-up.

PR #296, opened 2026-09-16
docs: clarify discovery 404 as identifier unknown (#274)

### LINKED ISSUE #274: 404 clarification
- 404 is overloaded. The new error-handling section defines 404 as "discovery endpoint not present", while the unchanged text further down and the OpenAPI spec define a `/discovery` 404 as "TEI unknown". A client can't tell "fail over to the next endpoint" from "stop" -> I would recommend sticking with TEI unknown. Otherwise, let's discuss.

Mentioned by @taleodor in #261

Mehrn0ush:
## Summary
- TEA `/discovery` `404` with an error body means TEI/PURL unknown — do not fail over solely for that
- Generic `404` without a TEA error body is a failed attempt / failover case
- `/discovery` is required on a conforming API base (unlike `/token`)

Closes #274


oej (approved):

oej:
Good work

PR #297, opened 2026-09-16
fix: clean up post-merge discovery 404 and checksum leftovers

Mehrn0ush:
## Summary
- Fix Tomcat Windows x64 `signatureUrl` in the `release` example (was plain `.zip.asc`)
- Raise `algValue` `minLength` to 40; MD5 was already removed in #264
- Define a TEA error body as `application/json` validating against `error-response`
- For discovery TEA `404`: stop, report identifier unknown, do not treat as “no updates”
- Remove redundant `404-discovery-identifier-not-found`; reuse shared object-not-found
Follow-up to #293 / #296 review notes from the `cd2fde4..main` scan.


taleodor:
Thanks! This closes out the #293/#296 notes cleanly (Windows x64 `signatureUrl`, TEA error body definition, terminal outcome for the discovery `404`, folding the duplicate 404 response). Two one-line leftovers I'd like fixed in this PR:

### 1. `/token` 404 wording still ambiguous

`discovery/readme.md` L468: "This is distinct from `/token`, where `404` **may** mean the token endpoint is not implemented."

`spec/openapi.yaml` (`/token` description): "A `404` from this endpoint means **only** that it is not implemented."

The readme's "may" reintroduces the "is it missing or not implemented?" ambiguity that #274 set out to remove. Suggest aligning to the OpenAPI wording:

> This is distinct from `/token`, where `404` means only that the token endpoint is not implemented.

### 2. `checksum` description lists an algorithm the enum no longer allows

`spec/openapi.yaml` L1627–1629 still reads "64 for SHA-256 / SHA3-256 / BLAKE2b-256 / **BLAKE3**", but #264 removed `BLAKE3` from the `checksum-type` enum. Same class of mismatch as the MD5 one this PR fixes via `minLength: 40`. Suggest:

> for example 40 for SHA-1, 64 for SHA-256 / SHA3-256 / BLAKE2b-256, 96 for SHA-384 / SHA3-384 / BLAKE2b-384, 128 for SHA-512 / SHA3-512 / BLAKE2b-512


Also note, that this PR likely will cause conflict with #270. I suggest merging this one first, then addressing 270 accordingly (that one has more issues).

Mehrn0ush:
Thanks @taleodor 
Fixed both. Aligned the discovery /token 404 sentence to “means only…”. Dropped BLAKE3 from the checksum length example; it was still in checksum-type on main, so removed it from the enum too so we don’t recreate the MD5-style mismatch. 
Agreed on merging #297 before rebasing #270.

Mehrn0ush:
One semantic thing I want to flag before merge

Discovery TEA `404` classification is still a bit broad.
We define a TEA error body (`application/json` validating as `error-response`) and say “do not fail over.” But the terminal wording currently treats that conforming `404` as "identifier unknown," which would also cover `OBJECT_NOT_SHAREABLE`. That can mis-report access denial / non-shareable as unknown, and it needs to stay coherent with [#270](https://github.com/CycloneDX/transparency-exchange-api/pull/270) (concealing `404` only after authentication).

Proposed split (same no-failover rule either way):
- `error: OBJECT_UNKNOWN` → stop; report "identifier unknown" to this authority
- other TEA `error-response` `404` (e.g. `OBJECT_NOT_SHAREABLE`) → stop; report "identifier unavailable to this client"
- non-TEA `404` → failed attempt / failover (unchanged)

@taleodor Does that split look right, or would you rather keep a single “unavailable” outcome for every TEA discovery 404?

taleodor:
Apparently Blake3 thing removal was wrong mb (contradicting things in different sources), should be restored in enum and in the length prose, the rest is good. + We should wait for #270 merge before proceeding with this one in any case.

Mehrn0ush:
Rebased onto main (#270 in). 
Mechanics look done. 
@taleodor 
Still want a call on discovery TEA 404 reporting: split OBJECT_UNKNOWN → “unknown” vs other TEA errors (e.g. OBJECT_NOT_SHAREABLE) → “unavailable” (no failover either way), or one outcome for every TEA discovery 404? 
And is sharing 404-object-by-id-not-found with #270’s concealing-404 prose OK, or should discovery have its own response again?

taleodor:
> Rebased onto main (#270 in). Mechanics look done. @taleodor Still want a call on discovery TEA 404 reporting: split OBJECT_UNKNOWN → “unknown” vs other TEA errors (e.g. OBJECT_NOT_SHAREABLE) → “unavailable” (no failover either way), or one outcome for every TEA discovery 404? And is sharing 404-object-by-id-not-found with #270’s concealing-404 prose OK, or should discovery have its own response again?

Thank you!

1 - One client outcome. 404 at discovery is terminal (full stop) - do not fail over and don't report it as evidence of absence. The report should carry the error value as is and that's it. So please reword `report that the identifier is unknown` to `report that the identifier could not be resolved at this authority`, with the error code attached, and keep the rest as is.

Full list of changes requested here:

a. discovery/readme.md, lines 366-370
```
That conforming TEA `404` means this server does not resolve the identifier, whether
because it is unknown or because the server withholds it (see
`404-object-by-id-not-found`); the client shall not infer which from the status alone.
The client shall not treat it as “`/discovery` is missing” and shall not fail over to
another endpoint solely because of it. The client shall stop discovery for that
identifier at this authority and report that the identifier could not be resolved
there, including the `error` value received. That outcome shall not be reported as
evidence that no updates are available.
```

b. discovery/readme.md, lines 462-465
```
- On `/discovery`, a TEA error response (`application/json` body with a string `error`
  property, typically `OBJECT_UNKNOWN`): this server does not resolve the TEI or PURL,
  whether unknown or withheld. Do not fail over. Stop and report that the identifier
  could not be resolved at this authority, with the `error` value; do not report that
  as evidence that no updates are available.
```

c. spec/openapi.yaml, /discovery description around line 910
```
and means this server does not resolve
the identifier, not that `/discovery` is optional or unimplemented.
```

d. spec/openapi.yaml lines 2438-2439
```
Body is a TEA `error-response` (for `/discovery`, typically
`error: OBJECT_UNKNOWN`).
```

2 - Keep it shared. Per #270 we don't disclose which case it is, if we split those, we break its premise.


Additionally, let's make error handling more generic, more changes (already reflected in b. above):

e. discovery/readme.md lines 360-364
```
A response is a
TEA error response only when its `Content-Type` is `application/json` (optionally with
parameters such as `charset`) and the body is a JSON object with a string `error`
property (typically `OBJECT_UNKNOWN`). Clients shall ignore properties they do not
recognize and shall not reject the response for an `error` value they do not know, so a
later TEA version can extend `error-response` without turning its `404`s into failover
triggers. Other JSON 404 bodies (for example `{"message":"Not Found"}`) are not TEA
error responses.
```

f. discovery/readme.md lines 381-382
```
- A `404` that is not a TEA error response as defined above (for example a web server
  answering for an unmounted path, or a JSON body without an `error` property)
```

g. spec/openapi.yaml lines 1791-1794, error-response description
```
TEA API error body. Servers shall emit exactly this shape. Clients recognize a TEA
error by an `application/json` response whose body is a JSON object with a string
`error` property; they ignore unrecognized properties and tolerate `error` values
they do not know, so that later versions can extend this schema. Other JSON 404
bodies are not TEA error responses.
```

Mehrn0ush:
Thanks — applied.

Mehrn0ush:
One leftover lead-in still reads “If the TEI (or PURL…) is not known to the TEA server, the discovery endpoint shall return 404…”, while the next paragraph now covers unknown or withheld. Want that opener reworded to match (e.g. “If this server does not resolve the TEI/PURL…”), or leave it? @taleodor

taleodor:
Yes, please reword it, and while you are there fix one more line I pasted wrong earlier.

1. 
```
If this server does not resolve the TEI (or PURL, when discovering by PURL), whether
because it is unknown or because the server withholds it, the discovery endpoint
shall return `404` with a TEA error response body.
```

2. spec/openapi.yaml lines 2439 to 2440, the shared 404-object-by-id-not-found response (my earlier d.)
```
Body is a TEA `error-response`, typically `error: OBJECT_UNKNOWN`.
```

Thank you!

Mehrn0ush:
Done

taleodor (approved):

PR #300, opened 2026-09-17
docs: clarify latest revision and in-place pre-signed URLs

Mehrn0ush:
## Summary
- Define collection/artifact \`latest\` as the greatest published \`version\`
- Clarify latest artifact download/signature resolve the same way (not circular)
- Note that a renewed pre-signed external \`url\` may be replaced in place without a new revision
Follow-up to #294 / #287 review: wording only.

## Test plan
- [ ] Skim OpenAPI \`/latest\` descriptions and artifact \`url\` pre-signed paragraph


Mehrn0ush:
@taleodor Could you take a look when you have a moment and confirm this matches what you had in mind?

taleodor (approved):

PR #301, opened 2026-09-17
docs: use lowercase ISO shall/should/may wording (#299)

### LINKED ISSUE #299: ISO grammar: shall, should, etc
Should be lower-case shall, should, etc. Needs thorough cleanup, preferably after other PRs are settled.

Mehrn0ush:
## Summary
- Use lowercase ISO requirement verbs in six documentation files: `api-flow/consumer.md`, `auth/readme.md`, `discovery/readme.md`, `signatures/signature.md`, `tea-product/tea-product.md`, and `tea-product/tea-product-release.md`.
- `spec/openapi.yaml` and `archive/` are unchanged.
- `doc/tea-requirements.md` is excluded until its normative scope is decided (#339, #343). Leaving that file unchanged defers the "needs to" wording tracked in #343, so #343 stays open.

Closes #299


taleodor:
let's postpone this for now and redo after everything else merges, so we don't get back to this and don't introduce unnecessary conflicts

api-flow/consumer.md
```diff
@@ -19,8 +19,7 @@ The standard TEI points to a product release. A product release is something sol
   can belong to multiple Component or Product Releases.
 - __List of artefact formats__: An artefact can be published in multiple formats.
 
-The user has to know product release TEI and in some cases version of each component (TEA 
-Component Release) to find the list of artefacts for the particular Product Release.
+Finding the list of artefacts for a Product Release starts from the product release TEI and, in some cases, the version of each component (TEA Component Release).
```
taleodor: Please keep line length constraints.

taleodor (changes requested):
Where "needs to"/"has to" became "shall", though, a few new requirements appear (remove requirements or change to should based on intent):

1. **doc/tea-requirements.md** is the first chapter of the Ecma draft, and after this PR it
   states in ISO terms requirements TEA 1.0 does not meet: well-known URLs "shall point to"
   a CLE document, version list and vendor discovery (`.well-known/tea` has only
   `schemaVersion` and `endpoints`); a repository URL "shall be found" from PURL / EAN /
   SKU / hash (discovery is TEI-only); the API "shall provide a way to publish an
   artefact" and "to search the inventory" (the API is read-only apart from `/token`); and
   "Authn/Authz shall be supported" while auth/readme.md allows servers with no
   authentication. As your commit message notes, the file's normative status is open
   (#339, #343). Could you leave this file out of the PR until that is decided, rather than
   defaulting it to "shall"?
2. **discovery/readme.md**: "The vendor shall ensure that the TEI is unique" contradicts the
   later text supporting one identifier (e.g. an EAN) resolving to several product releases,
   where separate TEIs are only a "should". Suggest "should", or say what uniqueness covers.
   The domain-name rewrite also drops that the vendor registers/controls the domain, which
   is what makes a TEI unique across vendors; worth keeping ("a DNS name under the vendor's
   control").
3. **signatures/signature.md**: "shall protect the private signing key" is unverifiable and
   about key management, which the spec leaves out of scope. Descriptive wording fits this
   background section: "The private key is used for signing and is kept protected."
4. **tea-product.md**: "the API might describe the format of the version string" keeps an
   old design question in the standard; the API has no such field. Suggest deleting it, or
   stating that the specification does not constrain version string format.

signatures/signature.md
```diff
@@ -18,19 +18,16 @@ signature files, a detached signature.
 
 Digital signatures provide integrity and identity to published data.
 
-- __Integrity__: Documents dowloaded must be the same
-  as documents published
-- __Identity__: Customers need to be able to verify the
-  publisher of the documents and verify that it is
-  the expected publisher.
+- __Integrity__: Signature verification enables detection of changes to signed documents.
+- __Identity__: Signature verification, together with an appropriate trust mechanism, enables verification of the signer's identity.
```
ppkarwasz: This file will not be in the spec, so we can omit it from the PR.

oej (approved):
Approved by meeting

Issue #285, opened 2026-09-14
Expand error-response coverage for 406, signatures, /token 404, and 400/204 cleanup


Mehrn0ush:
## Summary

Several documented HTTP outcomes cannot be expressed accurately with the current `error-response` model. `unknown-error-type` only allows `OBJECT_UNKNOWN` and `OBJECT_NOT_SHAREABLE`, but download flows also need to signal “no matching format” and “no signature,” `/token` documents an unimplemented-endpoint `404` that OpenAPI does not declare, generic `400` bodies are unspecified, and an unused body-bearing `204` component remains in the spec.

#270 improves concealment (`403` vs concealing `404`) but does not close these gaps. This issue tracks the remaining error-coverage work as a follow-up after #270 merges.

### 1. `406` cannot name “format unavailable”
Download operations declare `406` via `406-no-acceptable-format`, whose body is `error-response` → `unknown-error-type`. That enum only has:
- `OBJECT_UNKNOWN`
- `OBJECT_NOT_SHAREABLE`

Neither means “the revision exists but no format matches `mediaType` / `Accept`.”

### 2. Missing signature uses a generic `404`
Artifact signature download prose returns `404` when the selected format has no published signature, but the response is the same `404-object-by-id-not-found` / object-unknown path used for missing resources. Clients cannot distinguish “artifact unknown” from “signature not published.”

### 3. `/token` `404` is documented but not in OpenAPI
`/token` description says a `404` means the endpoint is not implemented (open / no-auth servers). Declared responses are only `200` / `400` / `401`. The operation table in published docs therefore omits a documented outcome.

### 4. Generic `400` body is unspecified
`400-invalid-request` uses `application/json: {}` with no schema. Either define a common structured body, or state explicitly that clients must not depend on `400` contents (token errors already have `token-error-response`).

### 5. Unused `204-common-delete`
`components.responses.204-common-delete` declares JSON content on a bodyless `204`, and nothing `$ref`s it. It should be removed or corrected if delete is ever in scope.

## Proposed direction

- Extend `unknown-error-type` (or split error enums by context) with distinct codes such as:
  - `NO_ACCEPTABLE_FORMAT` for `406`
  - `SIGNATURE_NOT_FOUND` for signature-absent `404`
- Keep concealment semantics from #270: concealing `404` remains `OBJECT_UNKNOWN` / not-shareable as already described there — do not overload those for format/signature cases.
[...]

PR #303, opened 2026-09-17
fix: name CLE as TEA projection and disambiguate CLE 404 (#278)

### LINKED ISSUE #278: CLE in OpenAPI vs ECMA-428 CLE 1.0.0 — align or name a TEA projection?
## Summary

TEA’s OpenAPI CLE schemas are described as based on ECMA-428 TC54 TG3 CLE Specification v1.0.0, but the model does not match the pinned CLE 1.0.0 schema ([cle-1.0.0.schema.json](https://tc54.org/schemas/cle/cle-1.0.0.schema.json)). As written, a payload that validates against TEA OpenAPI is not necessarily a valid CLE 1.0.0 document, and vice versa.

Before patching individual fields, it would help to decide whether TEA 1.0 intends CLE responses to be CLE documents, or a TEA-specific projection with explicit conversion rules.

Relevant schemas: `cle-event-type`, `cle-version-specifier`, `cle-event`, `cle-definitions`, and `cle`.

## Current gaps:

Compared with CLE 1.0.0:

- Incomplete envelope: CLE requires top-level `$schema`, `identifier`, and `updatedAt` (plus `events`). TEA’s `cle` schema only requires `events`.
- Identifier shape: CLE `componentRenamed` identifiers use `type` / `value`. TEA reuses its own `identifier` shape (`idType` / `idValue`).
- Version selectors: CLE version entries require `range`. TEA’s `cle-version-specifier` also allows `{version: ...}`, and examples use both forms.
- Event-specific requirements: CLE makes fields required by event type (e.g. `released` → `version`, `withdrawn` → `eventId`, end-of-* → `versions`, etc.). TEA lists those fields as optional on a single flat `cle-event`, so incomplete events can still validate.
- Unavailable CLE: README says CLE inclusion is optional, but the response contract when CLE is not available for a product/component/release is not clearly defined in OpenAPI.

Adding one missing field (for example only `updatedAt`) would patch a symptom and still leave a half-specified "CLE" surface?

## Decision:

A — Align with CLE 1.0.0: Treat TEA CLE endpoints as returning CLE documents. Reference or otherwise conform to the pinned CLE 1.0.0 schema, update TEA examples accordingly, and define behavior when CLE is unavailable.

B — Name a TEA-specific projection: Keep a TEA CLE-derived model in OpenAPI, but stop implying bit-identical CLE 1.0.0 documents. Explicitly name the projection and document conversion rules to/from CLE 1.0.0 (identifiers, version selectors, envelope fields, event-specific requirements).

## My question

For TEA 1.0, should CLE responses be CLE 1.0.0 documents (A), or an explicitly named TEA projection with conversion rules (B)?

I notice I keep framing these as binary choices — brains apparently love an A/B. If there's an option C hiding in here, I'd love to hear it too.

Mehrn0ush:
Thanks @taleodor
## Summary
- Name \`cle\` as a TEA projection of ECMA-428 (not a CLE document), with mapping rules
- Clarify \`cle-version-specifier\` and \`componentRenamed\` identifiers
- Add \`NOT_IMPLEMENTED\` and \`404-cle-not-available\` so CLE-absent vs object-unknown share \`404\` but differ by error body
Closes #278  
Related: #285 (\`/token\` 404 body can adopt the same pattern later)


taleodor (changes requested):
One change: keep a single 404 component. #297 folded the discovery-specific 404 into the shared one so that a TEA 404 has one shape and the error value carries the distinction; a 404-cle-not-available component reopens that, and #285 would need a third for signatures. 

Please point the four CLE operations back at 404-object-by-id-not-found and give that component the two-bullet description with both examples (text below); the per-endpoint meaning already lives in the cle description. 

Also "shall be ordered by id descending" rather than the bare "Ordered", since it is a requirement.

Mehrn0ush:
To address @oej  feedback on #278, I added a short note to the README’s CLE section and the OpenAPI cle description: the endpoints remain a TEA projection, while a full CLE 1.0.0 document can be published either as a release-collection artifact or as a company-wide product compliance document.

README.md
```diff
@@ -85,7 +85,15 @@ Product lifecycle events are communicated through the
 [ECMA-428 Common Lifecycle Enumeration standard](https://ecma-international.org/publications-and-standards/standards/ecma-428/).
 This includes product rebranding, repackaging, mergers and acquisitions, and product milestone events such as end-of-life and end-of-support.
 
-Inclusion of CLE is optional and it may be introduced on the following levels:
+The TEA CLE endpoints return a TEA projection of CLE event content (see the OpenAPI
+`cle` schema), not a full CLE 1.0.0 document. Publishers that need a bit-identical CLE
+document may publish it as a TEA Artifact in a product or component release collection,
+or — when the document is company-wide rather than release-scoped — as a compliance
+document linked from the TEA Product. The same CLE artifact may be referenced from more
```
taleodor: "as a compliance document linked from the TEA Product" - this doesn't hold, there is no compliance type for this. 

pls change wording to:


Publishers that need a bit-identical CLE 1.0.0 document publish it as a TEA Artifact in a release collection, as artifact type OTHER until a dedicated type exists; the same artifact may be referenced from more than one collection. The CLE endpoints are not a substitute for that document.

taleodor (changes requested):

Mehrn0ush:
Thanks — updated

taleodor (approved):

PR #304, opened 2026-09-17
docs: clarify discovery cardinality, response shape, and PURL (#279)

### LINKED ISSUE #279: Clarify discovery cardinality, response array, and direct PURL lookup
## Summary

Discovery prose and OpenAPI disagree on how many product releases a TEI can resolve to. The discovery document also does not describe the actual success response shape, and it never documented the direct-PURL workflow that #265 added to OpenAPI (“spec only; documentation unchanged”).

## Current gaps

- `discovery/readme.md` says a TEI belongs to a single product release, then later allows the same identifier (e.g. EAN) to span multiple releases.
- Successful `/discovery` is not described as the OpenAPI array of `discovery-info` objects (`productReleaseUuid` + `servers[]`).
- OpenAPI allows `200` with an empty array; a no-match should be `404`.
- Direct `purl=` lookup (already-known API base; exactly one of `tei`/`purl`) is in OpenAPI but missing from the discovery narrative.
- `.well-known` endpoint list vs `servers[].rootUrl` should be explicitly related.

## Proposed direction

- A TEI MAY resolve to multiple product releases; vendors SHOULD prefer one.
- Successful lookup MUST return a non-empty array of `discovery-info`; no match MUST return `404` (not `200` + `[]`).
- Add `minItems: 1` on the discovery response array; document `productReleaseUuid` + `servers[]`.
- Document direct PURL discovery as requiring an already-known API endpoint.

This is consistent with the direction discussed in #274: on a conforming `/discovery` handler, 404 means the identifier is unknown (not “empty success”). #274 can stay focused on overloaded 404 / failover wording; this issue is about cardinality, response shape, and the #265 docs gap.

## My Question

Does this direction look right for a follow-up PR I'd put together?

Mehrn0ush:
Thanks @taleodor 
## Summary
- TEI may resolve to multiple releases; array order is priority
- Success is a non-empty \`discovery-info\` array (\`minItems: 1\`); no match is \`404\` + \`OBJECT_UNKNOWN\`
- Document well-known (where \`/discovery\` is called) vs \`servers[]\` (where the release is served); \`rootUrl\` need not be in well-known
- Example \`/discovery\` response; direct \`purl=\` requires an already-known API base
- Align well-known \`url\` “without trailing slash” with \`rootUrl\`
Closes #279


taleodor (changes requested):
Thank you. One change: pls keep a single 404 component. #297 folded the discovery-specific 404 into the shared one so that a TEA 404 has one shape and the error value carries the distinction. A 404-cle-not-available component reopens that, and #285 would need a third for signatures.

Please point the four CLE operations back at 404-object-by-id-not-found and give that component the two-bullet description with both examples (text below); the per-endpoint meaning already lives in the cle description. Also "shall be ordered by id descending" rather than the bare "Ordered", since it is a requirement.

Mehrn0ush:
Thanks — that CLE/shared-404 feedback applies to #303. It’s already addressed there in be9766f (four CLE ops → 404-object-by-id-not-found, two-bullet body with both examples, events shall be ordered by id descending). 

Do the cardinality / non-empty success array / .well-known vs servers[] / direct purl= changes look like they’re heading in the right direction?

taleodor:
Yes, right direction on all four.

Edits

1. Finish the handoff sentence. The servers[] paragraph says how to select a server but not what happens next. Add: "All further requests for that product release are built as rootUrl + /v + the selected version + path. The API base used for /discovery is not used again for that release unless it is also listed in servers[]." That is the sentence that stops a client from continuing on the well-known base.

2. Align the no-match wording with #297. Both the readme and the /discovery description say "if nothing matches, 404 with OBJECT_UNKNOWN". The paragraph directly below, from #297, says the 404 means the server does not resolve the identifier, unknown or withheld, and the client shall not infer which. Say "if the server does not resolve the identifier" in the new sentences, so the two paragraphs describe one rule.

3. Dead TOC link. The TOC keeps "The TEA Version Index", and no such heading exists in the file. The PR edits the TOC, so drop that entry while there.

taleodor (approved):

PR #305, opened 2026-09-17
Fix/error response coverage

### LINKED ISSUE #285: Expand error-response coverage for 406, signatures, /token 404, and 400/204 cleanup

Mehrn0ush:
## Summary
- Extend \`unknown-error-type\`: \`NO_ACCEPTABLE_FORMAT\`, \`SIGNATURE_NOT_FOUND\`, \`INVALID_REQUEST\`, \`INVALID_PAGE_TOKEN\` (keeps \`NOT_IMPLEMENTED\` from #278)
- Signature \`404\` and \`406\` name distinct codes; concealment still uses \`OBJECT_UNKNOWN\` for sub-resources
- Declare \`/token\` \`404\` with no content requirement
- Resource \`400\`/\`401\`/\`403\` may carry \`error-response\`; clients shall not require a body
- Remove unused \`204-common-delete\`

Closes #285  

Depends on #303 (#278) for shared enum / CLE \`NOT_IMPLEMENTED\`

## Test plan
- [ ] Enum lists all seven values once
- [ ] Both signature downloads use \`404-artifact-signature\`
- [ ] \`/token\` has \`404-token-not-implemented\` with no \`content\`
- [ ] No \`204-common-delete\`

Thanks @taleodor


taleodor (changes requested):
Similar to my response for #303 please fold all 404s into single existing 404-object-by-id-not-found.

Mehrn0ush:
Thanks.
done. 
One question - should /token be the one carve-out from the shared TEA 404 shape (status 404, body not required / not meaningful), while CLE / signature / discovery keep required TEA error bodies on the shared component? That matches how discovery already treats “TEA 404 with body” vs “generic web 404.”

If yes, I’ll restore a dedicated /token 404 response, keep resource 404/406 bodies required for the structured cases, and limit "clients shall not require a body" to 400/401/403 only.

taleodor:
Thank you, no carve-out; what you have now is right. /token uses the shared 404 with NOT_IMPLEMENTED, and please add one sentence there: a server that does not implement the endpoint shall still answer with that body, a 404 without a TEA error body is not a conforming answer.

Two small things: 
- the latest-signature endpoint needs the same SIGNATURE_NOT_FOUND note as the versioned one, and 
- the "any 4xx may carry an error-response" sentence should live once on the error-response schema rather than in three response descriptions. 

Lands after #303.

Mehrn0ush:
Thanks. 
done

taleodor (approved):

PR #306, opened 2026-09-17
fix: require a collection for every release (#286)

### LINKED ISSUE #286: Release without a collection — optional latestCollection or always require one?
## Summary

Supporting docs describe collections as optional for a Component Release, but the OpenAPI response model for a component release with collection requires `latestCollection`. A release that has no collection yet therefore has no defined successful representation.

This is separate from #264’s decision to allow `artifacts: []` on a collection that exists. Empty collection ≠ missing collection.

## Current gaps

- `component-release-with-collection` requires both `release` and `latestCollection`.
- README states each Component Release *may* have its own TEA Collection.
- There is no stated successful representation for GET of a release that has never published a collection, or for list endpoints when none exists.
- The dedicated “latest collection” endpoint reuses the same generic `404-object-by-id-not-found` for both “release doesn’t exist” and “release exists but has no collection” — ambiguous, not distinguishable by clients (same pattern as signature-absent `404`).

## Decision needed

- A — Keep collections optional:
Require only `release` on the combined response; omit `latestCollection` when none exists. Collection list endpoints may return an empty result. When no collection has been published, the dedicated latest-collection endpoint still returns `404`, but that outcome should be distinguishable from “release not found” (e.g. a distinct error code, similar to the signature-absent `404` discussion). Align OpenAPI `required` and docs with that model.

- B — Require a collection for every release:
Every Component Release (and, if the same rule applies, Product Release) always has a collection, which may be empty (`artifacts: []` per #264). Keep `latestCollection` required. Update README and related docs that currently say collections are optional.

Half-fixing (e.g. only dropping `required` without endpoint behavior) leaves clients without an interoperable “no collection” path?

## Relation to other work

- #264 / #259: required fields and empty `artifacts: []` — complementary.
- Does not change artifact download, auth, or discovery issues.

## My question

For TEA 1.0, should a release without a collection be representable (A), or must every release always have a collection (B)?

Mehrn0ush:
## Summary
- Every component/product release has a TEA Collection (may be empty \`artifacts: []\`)
- Empty first version uses \`INITIAL_RELEASE\`; first artifacts publish as version 2
- Make \`artifacts\` required on \`collection\`; \`latestCollection\` stays required
- \`/collection/latest\` \`404\` means the release is unknown
- Align collection docs: \`createdDate\`, \`COMPONENT_RELEASE\` / \`PRODUCT_RELEASE\`

## Test plan
-  Skim README, tea-requirements, tea-collection.md, OpenAPI \`collection\` + both \`/collection/latest\`

Thanks @taleodor

Closes #286


taleodor (changes requested):
Two edits and this is good. (1) The component-release-with-collection example is now inconsistent with the rule: version: 2 with INITIAL_RELEASE and artifacts. Make it version: 1, or keep 2 and use ARTIFACT_ADDED. (2) In tea-collection.md, replace the quoted "version 1 is whatever was first observable" line with plain prose, e.g. "In both cases version 1 is the first collection a client could have retrieved." While in that list, the empty bullet above artifacts can go.

Also, Line 333 is a stray empty bullet ( -) between updateReason and artifacts. Pre-existing, but the list is being edited in this PR, so pls correct.

Mehrn0ush:
Thanks.
Fixed both: example now uses ARTIFACT_ADDED at version 2 with artifacts; tea-collection.md uses the plain "first collection a client could have retrieved" wording and drops the empty bullet.

taleodor (approved):

PR #307, opened 2026-09-17
fix: tighten TEA API base URLs and TLS identity (#288)

### LINKED ISSUE #288: Tighten discovery/API base URL, TLS identity, and credential boundaries
## Summary

`.well-known` `endpoints[].url` and `tea-server-info.rootUrl` are still `format: uri` with no HTTPS / component restrictions. That undercuts #276’s TLS and credential-boundary rules if a conforming base can be `http://`, and it makes `/v{version}/…` appending ambiguous (userinfo, query, fragment, trailing slash).

Follow-up to #276 / #275. Related but separate from #282. Independent of #261 TEI syntax.

## Proposed

Require absolute HTTPS base URLs with no userinfo, query, fragment, or trailing slash (well-known `url` and discovery `rootUrl`).

Optional: explicit hostname checks per [RFC 6125 section 6](https://www.rfc-editor.org/rfc/rfc6125.html#section-6); same “no TEA token to external URLs” sentence on `release-distribution.url` / `signatureUrl` (already stated for artifact formats / redirects).

## My question

For 1.0: should absolute HTTPS base URLs be required?
If yes, include the TLS-identity and `release-distribution` riders in the same PR, defer them, or drop as non-blocking?

Mehrn0ush:
## Summary
- Pattern on well-known \`url\` and \`rootUrl\`: absolute, no userinfo/query/fragment/trailing slash (\`http\` allowed in schema for local test docs)
- Conforming deployments shall advertise only \`https\`; clients reject \`http\` unless configured for local testing
- Cite RFC 9525 for server identity checks; clarify OpenAPI \`servers\` local-dev example
- Copy no-TEA-token-to-external-URL wording onto \`release-distribution\` \`url\` / \`signatureUrl\`

Closes #288

Thanks @taleodor


taleodor (changes requested):
Validated: spec passes, pattern identical in both schemas, well-known and readme examples still validate, RFC 9525 cited in all four places, `release-distribution` riders and the `servers` label are in. 

Pls address following:

1. The local-testing carve-out is missing its guard. Append to the new paragraph in `discovery/readme.md`:
 ```
A client shall reject an `http` base URL unless it has been explicitly configured to
allow that specific base for local testing. The allowance is per configured base URL,
never a global setting, so it cannot apply to a base the client learned from discovery.
Credentials may be sent to a base allowed this way; a deployment that relies on it is
not conforming.
```

2. `signatures/signature.md` now cites RFC 9525 directly above a bullet that still says "(CN or Subject Alt Name)". RFC 9525 §2 excludes the Common Name from identity checks; only SANs count. Drop "CN or".
3. The pattern's path class `[^?#]*` admits whitespace (`https://host/a b` matches). `[^?#\s]*` closes it; `format: uri` does not, since validators treat `format` as annotation.
4. The scheme match is lowercase-only, which is the normalized form and fine; the description could say "lowercase `https`" so nobody reads the rejection of `HTTPS://` as a bug.
5. In `discovery/readme.md` the touched sentences were lowercased to "shall" while the surrounding "SHALL NOT" sentences in the same block were not, so the block is now mixed. Pls use lowercase everywhere.

P.s. edited due to earlier posting error

Mehrn0ush:
Thanks.
Addressed all five: per-base local-test guard, SAN-only identity bullet, \s excluded from the path pattern, lowercase https called out in the descriptions, and lowercase shall/should/may in the mixed discovery auth blocks.

taleodor (approved):

PR #315, opened 2026-09-20
Add more clear definition of the TEI:

oej:
Copied almost verbatim from the VERS specification


ppkarwasz:
TEI types have no other purpose than validation, so we might as well have them in a registry in this repository, instead of including them in the specification that needs to be amended by ECMA.

Both PURL and VERS use an external registry at https://packageurl.org/

oej:
I kind of agree, the only problem is lack of time. If we move to a separate registry we need a process to maintain the types. For now, we can make clear that the types in the spec is not normative, they are examples and during the process towards ECMA standardisation we have time to set up our own registry and process. OK?

ppkarwasz:
Sounds good to me, but I'd double-check with @pombredanne and @stevespringett to see if we can add such phrasing to the standard.

mjherzog:
From our PURL and VERS experience, the structured definition of **types** is  the first step - in both cases a JSON schema. 
This provides the capability to minimally validate a TEI for syntactic validity and this worked well for PURL long before we added **type** registration. The key decision is whether such a TEI **type** schema is normative or informative for this cycle. In either case it should be very useful for adoption.

The  idea of registering **types** is a second step and requires the previous definition of the **type** records. I would recommend focusing on the definition of a simple v1 schema for TEI **types** and leave the registry to a later standards cycle.

mjherzog:
For the example:
`tei://cyclonedx.org/purl/cGtnOnB5cGkvY3ljbG9uZWR4LXB5dGhvbi1saWJAOC40LjA_ZXh0ZW5zaW9uPXdobCZxdWFsaWZpZXI9cHkzLW5vbmUtYW55`
Is `cGtnOnB5cGkvY3ljbG9uZWR4LXB5dGhvbi1saWJAOC40LjA_ZXh0ZW5zaW9uPXdobCZxdWFsaWZpZXI9cHkzLW5vbmUtYW55`
the BASE64URL encoding of the PURL:
`pkg:pypi/cyclonedx-python-lib@8.4.0?extension=whl&qualifier=py3-none-any` ?
This is implied but should probably be stated explicitly.

This raises the question of how to document the encoding algorithm in the schema.

oej:
Rebased branch and added clarification of PURL encoding.

oej:
It is important to remember that we are just creating unique identifiers based on existing IDs - not carrying any data. The TEI should be simple to create based on existing identifiers. For most commercial products, it likely won't be a PURL but something else.

ppkarwasz:
For this version, should we restrict the type-specific part to **unreserved** characters of [RFC 3986](https://www.rfc-editor.org/info/rfc3986/), i.e. ASCII letters, digits, `-`, `_`, `~` and `.`?

I think we should define the encoding of PURL as Base64URL encoding of the **canonical** for of PURL **without** padding.

taleodor:
Most of this has since landed elsewhere: the domain-name wording (#301), PURL as unpadded Base64URL of the canonical PURL (#370), the lowercase scheme and type as part of the canonical form (#371), and #373 moves the types into a registry. Two parts are still worth keeping, and I'd suggest folding them into #373 and closing this:

1. The type-name syntax, in Discovery's normative TEI syntax (clients have to parse unknown types). The WG needs to choose between this PR's rule (letters, digits, `.`, `-`, starting with a letter) and #373's (lowercase letters and digits only). Either way, drop "case insensitive": since #371, TEIs compare case-sensitively and the type is lowercase.
2. "A TEI is a URI as defined in RFC 3986", which `main` doesn't say anywhere yet (the current wording "a vlid URI scheme" needs fixing).

PR #326, opened 2026-09-21
feat: align checksum algorithms with CycloneDX 1.7

ppkarwasz:
## Align checksum algorithms with CycloneDX (ECMA-424)

This PR adds `Streebog-256` and `Streebog-512` to the `checksum-type` enumeration, mirroring the two hash algorithms introduced in CycloneDX 1.7 by CycloneDX/specification#525.

A second commit restores `MD5`, which was dropped in #264 without discussion. CycloneDX 1.7 still lists it, and publishers may only have MD5 digests available for legacy artifacts. With both commits the enumeration is identical to CycloneDX's `hash-alg`, and the `algValue` minimum length is lowered to 32 hex characters to match the MD5 digest size.

A third commit replaces the inline enumeration with a `$ref` to the `hash-alg` definition of the published CycloneDX 1.7 schema, so the two cannot drift apart. Note that tools reading the OpenAPI document now need network access to cyclonedx.org to resolve the reference; vendoring the schema instead would also require vendoring its three sibling schemas and a change to the ECMA build, so that is left for a separate discussion.

Following the discussion below, the fourth commit settles what is normative and what is not. The normative source of the algorithm names is the `hash-alg` definition of ECMA-424, undated, so a later edition of ECMA-424 can add algorithms without a TEA release; the `$ref` to the 1.7 schema stays as an informative snapshot that gives generated clients a typed enum. Clients shall ignore checksums whose algorithm they do not recognise. To make that hold for the generated clients too, the four generator configurations set `enumUnknownDefaultCase: true`, which maps a value outside the snapshot onto an `unknown_default_open_api` member instead of failing to deserialize the whole response; this also protects every other enum the specification may extend later. ECMA-424 is already an undated normative reference of the Ecma document, so nothing changes on that side.

Since the `checksum` description needed reflowing anyway, and OpenAPI descriptions are interpreted as Markdown, it now uses [semantic line breaks](https://sembr.org). The rendered output is unchanged in structure, and future edits to the description will produce shorter diffs. The digest-size examples are also turned into a list to make room for the new algorithms.

Verified with `openapi-generator-cli v7.24.0` (the version used by CI): `validate`, full client generation for all four languages, and by rendering the schema through the ECMA-xxx-TEA converter. The Java and Python `ChecksumType` enums contain the CycloneDX values plus the unknown-value member.


ppkarwasz:
We also lack `MD5`, which is listed in the CycloneDX specification. Did we omit it on purpose?

ppkarwasz:
MD5 was removed in #264. I'd propose keeping it as a supported checksum algorithm, for three reasons:

- **Integrity**: MD5 is still a perfectly adequate checksum for verifying that a download completed correctly, and for many legacy systems it's the only checksum algorithm available.
- **Authenticity**: MD5 doesn't provide this, but neither do the other checksum algorithms on their own. For artifacts self-served by the TEA server, MD5 is no weaker than any other option in this respect.
- **Alignment with CycloneDX**: MD5 is one of the algorithms supported by CycloneDX 1.7. The current CycloneDX 2.0 draft has not performed any changes to the list of supported hash algorithms.

oej:
If we re-add it, we have to be very clear that it is only for legacy support. We should not encourage any use of MD5.

ppkarwasz:
That warning should probably end up in the unfinished `signatures/signature.md`. Do we expect that file to end up in the normative part?

Honestly, I wouldn't demonize MD5, the same way nobody demonizes CRC32 being used in the Zip format. `algValue` is a *check sum*: it detects a corrupted download and provides no protection against malicious modification, whatever the algorithm. There is slight protection for artifacts hosted elsewhere, assuming the TEA server is not compromised, but SHA-512 doesn't remove that assumption either. Trusting the TLS connection with the server to authenticate API responses shouldn't be enough anyway.

What counts is the hash used by the signature algorithm, and the API doesn't return it: `signatureUrl` explicitly "does not define which signature technology is used, nor model the signing algorithm, key, or certificate chain". Restricting `checksum-type` would suggest otherwise. So in `checksums` MD5 poses no problem; it does in the elements the API treats as opaque, detached signatures and the PKI / GPG web of trust behind them. That guidance belongs under "External (detached) digital signatures" in `signature.md`, where `indication of hash algorithm` is still an empty bullet. Happy to draft it separately.

taleodor:
PR lists MD5 in prose, but doesn't add it. I don't think we should merge this currently, I think MD5 can be retired and we can live without exotic algorithms for 1.0.0. We should be able to revisit after.

ppkarwasz:
979e428 does add the enum value, not only the prose.

Honestly I don't think it is up to TEA to decide which checksum algorithms may appear in the metadata. We transport digests publishers already have: a legacy artifact may only have MD5, and a publisher under Russian requirements will have Streebog. If the value is missing from the enum, the server cannot describe an artifact it holds. Revisiting after 1.0.0 is also costly, since it requires a new version of the standard.

MD5 and SHA1 are common on Maven Central and taken together, they provide a pretty decent collision protection.

My proposal: align the names with CycloneDX, the closest standard and the language of TEA answers once 2.0 lands, and add a note to `checksum-type` that the list is not normative and the normative one is CycloneDX `hash-alg`.

The IANA registries ([RFC 6920](https://www.iana.org/assignments/named-information/named-information.xhtml), [RFC 8122](https://www.iana.org/assignments/hash-function-text-names), [RFC 9530](https://www.iana.org/assignments/http-digest-hash-alg/)) serve different purposes and none uses our names.

oej:
Having a list somewhere helps developers and promotes interoperability, but it may also hinder future additions. We either skip having our own list and point to something else, maintain the list externally (like we wanted to do with TEI types or let it be a free text and let the developers run crazy and get into interoperability problems. What do we think is the best way forward? Having to update the ECMA standard to change this list may not be a good thing.

ppkarwasz:
In a16603064bf6e97cc29480e724f1e9ea76f618b5 I replaced the list with a reference to CycloneDX 1.7.

We can also say that the naming of hash algorithms is given by the CycloneDX standard and implementations should skip checksums they don't understand.

oej:
What is the change process of CycloneDX for this? Before we connect the standard that close we need to know if it changes for the positive or if they have the same issue as we would have.

ppkarwasz:
I don't know, but they are bound by backward compatibility too.

This change is dictated by my “maintainer mode”, which recommends pushing problems onto others (as long as I trust them, which in the case of the CycloneDX I clearly do).

oej:
If we do this, we have to add the same reference as a normative reference in other parts of the docs.

ppkarwasz:
ECMA-424 is already a normative reference, so we can reuse it.

I update the PR to refer to CycloneDX without a specific version number and specify that the reference to CycloneDX 1.7 in the OpenAPI spec is a convenience for implementors

PR #332, opened 2026-09-21
UUID doc - small fixes

oej:
I don't remember if this is included in the ECMA spec or not, but anyway


doc/tea-uuid-scope.md
```diff
@@ -15,7 +15,8 @@ authoritative domain  +  object type  +  UUID
 
 The per-object-type scoping reflects how TEA servers are typically implemented: each object type lives in its own database table, where uniqueness is trivially enforced by a primary key. Cross-table uniqueness offers no functional benefit - TEA has no operation that resolves a UUID without already knowing its object type.
 
-Implementations MAY use generators that produce globally unique UUIDs (e.g. RFC 4122). Doing so neither violates nor strengthens this specification.
+Implementations MAY use generators that produce globally unique UUIDs (e.g. RFC 9562).
```
taleodor, line 18: MAY should be lowercased -> may.

taleodor (changes requested):

taleodor:
Can you replace last sentence with:

```
Implementations may use any UUID version defined in RFC 9562. However UUIDs are
produced, the exception above applies: a server shall not assign a UUID that is
already in use for the other release type.
```

Issue #313, opened 2026-09-19
Group endpoints and tags by resource in openapi.yaml


ppkarwasz:
## Problem

The `paths` block in `spec/openapi.yaml` is not grouped in a coherent way, which makes the spec harder to review and the generated documentation harder to navigate:

- The product section is scrambled: `/product/{uuid}/cle` and `/products` appear after the `/productReleases` endpoints, and the list endpoint (`/products`) comes after the single-item endpoint.
- The component side uses the opposite arrangement: `/components` comes before `/componentReleases`.
- The six collection endpoints alternate between component release and product release, so following a single resource means skipping every second entry.
- `/token` and `/discovery` are last in `paths`, while `TEA Authentication` is the first tag and `TEA Discovery` the second to last. Neither order follows the consumer flow.
- The three `token-*` schemas sit at the very top of `components.schemas`, ahead of the domain model.

## Proposal

Order `paths` one block per resource, in consumer-flow order. Within each block: list endpoint, single item, then sub-resources.

```
/discovery
/products
/product/{uuid}
/product/{uuid}/releases
/product/{uuid}/cle
/productReleases
/productRelease/{uuid}
/productRelease/{uuid}/cle
/productRelease/{uuid}/collections
/productRelease/{uuid}/collection/latest
/productRelease/{uuid}/collection/{collectionVersion}
/components
/component/{uuid}
/component/{uuid}/releases
/component/{uuid}/cle
/componentReleases
/componentRelease/{uuid}
/componentRelease/{uuid}/cle
/componentRelease/{uuid}/collections
/componentRelease/{uuid}/collection/latest
/componentRelease/{uuid}/collection/{collectionVersion}
/artifact/{uuid}/latest
/artifact/{uuid}/latest/download
/artifact/{uuid}/latest/signature/download
/artifact/{uuid}/{artifactVersion}
/artifact/{uuid}/{artifactVersion}/download
/artifact/{uuid}/{artifactVersion}/signature/download
/token
```

Order the top-level `tags` list the same way: Discovery, Product, Product Release, Component, Component Release, Artifact, CLE, Authentication.

Move the `token-*` schemas to the end of `components.schemas`, next to the other authentication-related components.

## Impact

Pure reordering. No change to the wire format, to generated clients, or to the rendered semantics of the spec.

oej:
I like this change, but as it is not normative we can make it after we send it in to ECMA for a review.

ppkarwasz:
[...]

Issue #311, opened 2026-09-19
Make internal component names consistently kebab-case in openapi.yaml


ppkarwasz:
## Problem

Everything that goes on the wire in `spec/openapi.yaml` is consistently camelCase: path segments, path parameters, query parameters and schema properties. The only exceptions are mandated by external standards and must stay as they are (OAuth 2.0 token fields and error codes in snake_case, CLE event types in camelCase, checksum algorithm names).

Internal component keys, which never appear on the wire, are kebab-case with a few outliers:

- `#/components/schemas/productRelease` is the only camelCase schema key. Its siblings are `product`, `component`, `component-ref`, `release-distribution`, `component-release-with-collection`, and so on.
- `#/components/securitySchemes/bearerAuth` and `basicAuth` are camelCase. This is the conventional OpenAPI spelling, but it does not match the rest of the file.
- In the `compliance-document-type` enum, `FedRAMP` is the only value not in SCREAMING_SNAKE case (compare `HITRUST`, `TISAX`, `CYBER_ESSENTIALS`).

## Proposal

- Rename the schema key `productRelease` to `product-release` and update all `$ref`s.
- Rename `release` to `component-release` and update all `$ref`s.
- Decide on `bearerAuth` / `basicAuth`: either keep the OpenAPI convention (and document the exception) or rename to `bearer-auth` / `basic-auth`, updating the top-level `security` block and any per-operation `security` entries.
- Decide on `FedRAMP`: keep the official brand spelling, or normalise to `FEDRAMP` to match the rest of the enum.

## Impact

The schema key rename is a pure refactor: generators already map `product-release` to `ProductRelease`, so generated class names do not change. The security scheme rename is also internal. Changing `FedRAMP` to `FEDRAMP` is a wire-format change for the compliance document identifier type and must land before 1.0 if done at all.

taleodor:
Implemented in #311, with three notes from doing it.

**`release` → `component-release` does change generated class names.** The impact section says the rename is a pure refactor because generators already map `product-release` to `ProductRelease`. That holds for `productRelease` — I verified the class name is unchanged — but not for `release`: the model becomes `ComponentRelease` in go, python, typescript and java-webclient. Still worth doing (the schema calls itself "A TEA Component Release", and `release` beside `component-release-with-collection` reads as two different things), but it is a source-compatibility break for anyone who generated a client on pre-released versions of TEA.

[...]

Issue #217, opened 2026-02-25
Clarify scope of all our UUIDs


oej:
We need to clarify the scope of each UUID - are they globally Unique in a TEA server implementation? If I move my TEA service from one server to another - should I preserve UUIDs?

taleodor:
I may be paranoid, but my approach is to have UUID unique within a single TEA server and within specific object scope (i.e. namespacing by components, component releases, etc - such that each object within the namespace is guaranteed to have unique UUID within the namespace).

This approach allows to have strong uniqueness guarantees via implementation side. I don't think we should pursue universal uniqueness, i.e. see here - https://alexsci.com/blog/uuid-oops/

taleodor:
@Mehrn0ush I'd appreciate that if you bring AI into conversation, you 

1) Give clear attribution to which parts are AI generated
2) Try to limit the size of the response

This is because we're perfectly capable to engage AI if something is not clear and we need to expand on points. Also the issue with AI here is that AI is trying to base concepts on its existing knowledgebase, while we're trying to build something new here.

Now, for your concerns, I'll expand:

TEA is a distributed system, there is no centralized authority that will or even can police UUIDs. If a malicious actor wants to create duplicate UUIDs, there is nothing to stop them.

Therefore, TEI contains "authoritative domain" part. Combination of authoritative domain + UUID is what creates global uniqueness, rather than UUID itself. For clarification, please read the discovery docs and watch any of my TEA talks on the TEI part, i.e. [@OWASP Global AppSec USA 2025](https://www.youtube.com/watch?v=lvEEMPNOEkI).

Now, why I want to further tighten scope to authoritative domain + object type + UUID - purely for practical implementation reasons. Usually, a single object type lives in its own table in the database. Enforcing UUID uniqueness is easy this way. If you try to enforce uniqueness across tables, this becomes much more difficult problem. And I believe there is no benefit in that, as we have no plans to resolve a UUID into object type + object. There is always expectation that you know object type alongside UUID. So practically, I believe this is the most reasonable approach. With that, if you want, you can still have globally unique UUIDs in your specific implementation - it won't break anything, but it also doesn't bring any real benefits.

oej:
Need to document this

oej:
Merged

PR #333, opened 2026-09-21
docs: forbid Product/Component Release UUID reuse (#329)

### LINKED ISSUE #329: Collection identity conflicts with permitted cross-type UUID reuse
## Summary

TEA allows the same UUID for a Product Release and a Component Release (`doc/tea-uuid-scope.md`), and a Collection inherits its release’s UUID. That yields two distinct Collection objects that share the same documented identity tuple, so clients caching by `(authority, object type, UUID, version)` can conflate them.

Verified on `main` @ `86356888` (after #325).

## Current gaps

1. **Cross-type UUID reuse is allowed**  
   [tea-uuid-scope.md](https://github.com/CycloneDX/transparency-exchange-api/blob/86356888a7fda3af1da347ea8bc1f1f62ee018fe/doc/tea-uuid-scope.md#L13-L14) requires uniqueness only within `(authoritative domain, object type)`. UUIDs need not be unique across object types, so a Product Release and a Component Release may share a UUID.

2. **Collection UUID is copied from the parent release**  
   [`collection.uuid`](https://github.com/CycloneDX/transparency-exchange-api/blob/86356888a7fda3af1da347ea8bc1f1f62ee018fe/spec/openapi.yaml#L1656-L1672) says the collection UUID matches the associated Product Release or Component Release UUID. `belongsTo` distinguishes the parent *kind*, but is not part of the uniqueness / identity rule.

3. **Concrete collision**  
   If Product Release `U` and Component Release `U` both exist under the same authority:
   - `GET /productRelease/{U}/collection/...` and `GET /componentRelease/{U}/collection/...` return two different Collection resources
   - both advertise `uuid: U` (same object type: Collection)
   - a client keyed on `(authority, Collection, U, version)` cannot tell them apart

Collections may be signed and cached. Ambiguous identity breaks verification, cache keys, and any “fetch collection by UUID” logic that does not also carry `belongsTo` or the parent path.

## Proposed direction

Pick one normative approach (or an equivalent):

1. **Widen collection identity** to include `belongsTo` (and/or parent resource type), and state that uniqueness for collections is `(authority, Collection, belongsTo, UUID)` (plus `version` for a revision); or
2. **Stop inheriting the release UUID** and give collections their own UUIDs, with a parent reference; or
3. **Forbid cross-type UUID reuse** for release types that share a collection namespace (narrower than (1), but may be enough).

Update `doc/tea-uuid-scope.md` and the `collection.uuid` prose together so identity and addressing stay aligned.

## Relation to other work

- Builds on the model from #217 / #245 (UUID scope).
- Distinct from API-base uniqueness (separate finding): this is about two Collections under one authority sharing one UUID.

Mehrn0ush:
## Summary
- Address the collection-identity collision from #329 by forbidding the same UUID for a Product Release and a Component Release under one authoritative domain (option 3).
- Prose only: `doc/tea-uuid-scope.md`, OpenAPI uuid descriptions, and `tea-collection` narrative. No schema/`required`/path changes.

Closes #329


doc/tea-uuid-scope.md
```diff
@@ -11,17 +11,19 @@ authoritative domain  +  object type  +  UUID
...
 
-Implementations MAY use generators that produce globally unique UUIDs (e.g. RFC 4122). Doing so neither violates nor strengthens this specification.
+The per-object-type scoping reflects how TEA servers are typically implemented: each object type lives in its own database table, where uniqueness is trivially enforced by a primary key. Cross-table uniqueness is not required in general (TEA has no operation that resolves a UUID without already knowing its object type), except that Product Release and Component Release UUIDs shall be disjoint as above, because those types share the Collection UUID namespace.
+
+Implementations may use generators that produce globally unique UUIDs (e.g. RFC 4122). Doing so satisfies the Product Release / Component Release rule in ordinary practice; deliberate reuse across those two types is what this exception forbids.
```
taleodor: RFC 4122 has been obsoleted by [RFC 9562](https://www.rfc-editor.org/info/rfc9562/)

taleodor (commented):
- Terminology in the line the PR edits. The object-type list still reads "(Product, Product Release, Component, Release, Collection, Artifact)" while the new paragraph says "Component Release"; the document header says "TEA Release". One name per type would help, and "Component Release" is the one the rest of the spec uses (the schema describes itself as "A TEA Component Release"; #311 proposes renaming the key to component-release).
- RFC 4122 → RFC 9562. The Implementations may use generators… line — which this PR touches — cites RFC 4122, obsoleted by RFC 9562 in 2024; openapi.yaml already says "A UUID in lower case (RFC 9562)". Further we probably don't need this reference at all - see below.
- "Doing so satisfies the Product Release / Component Release rule in ordinary practice" is a bit soft for a normative document. I would just say:

```
Implementations may use any UUID generator; the choice neither violates nor strengthens this specification. The exception above is a requirement on the server: it shall not assign a UUID that is already in use for the other release type, however UUIDs are produced.
```

taleodor:
The restriction is overly broad now. Can we take a step back and go for narrower fix from previous commit state?

See reasoning below:

1. It deletes the rationale from #217 without replacing it. The removed paragraph ("each object type lives in its own database table, where uniqueness is trivially enforced by a primary key; cross-table uniqueness offers no functional benefit — TEA has no operation that resolves a UUID without already knowing its object type") was the stated design basis of that document. Reversing it is a WG call, and the argument it made is still true: no TEA operation resolves a UUID without knowing its type.

2. It imposes a real implementation cost the narrow fix doesn't. Under per-type scoping, a primary key gives the guarantee for free. Under domain-wide uniqueness, "it shall not assign a UUID that is already in use" means a cross-table check on every insert (six lookups, or one shared allocator). Ironically this reintroduces the leeway you objected to earlier: a server generating random v4 per table cannot guarantee domain-wide uniqueness, only make collision negligible — so either the rule needs a check nobody is doing today, or it's satisfied probabilistically, which is what the wording change was meant to avoid.

3. With narrow fix, collection UUIDs are exactly the union of Product Release and Component Release UUIDs, so PR ∩ CR = ∅ is sufficient for unique collection identity. Product-vs-Artifact reuse never caused the collision.

Mehrn0ush:
Restored per-type UUID scoping and its rationale, limiting the additional uniqueness requirement to Product Release and Component Release. This prevents the Collection identity collision without requiring uniqueness across unrelated types. All three documents now distinguish Collections  belonging to different releases from versions of the same Collection. The terminology, RFC-reference removal, and ISO normative wording are retained.

ppkarwasz:
> The restriction is overly broad now. Can we take a step back and go for narrower fix from previous commit state?

We are talking about 128-bit numbers. If properly generated, all the TEA servers in the world can be merged into a single one, without any collision. Do we really need to have separate namespaces for different objects?

taleodor:
> > The restriction is overly broad now. Can we take a step back and go for narrower fix from previous commit state?
> 
> We are talking about 128-bit numbers. If properly generated, all the TEA servers in the world can be merged into a single one, without any collision. Do we really need to have separate namespaces for different objects?

We had this conversation already. In practice collisions happen more frequently than in theory, for various reasons. The issue on the implementation side is that it's easy to enforce no collision rule within a single table, but more difficult across tables.

taleodor (approved):

doc/tea-uuid-scope.md
```diff
@@ -11,17 +11,19 @@ authoritative domain  +  object type  +  UUID
 ```
 
 - The **authoritative domain** is supplied by the TEI (`tei://<domain-name>/<type>/<unique-identifier>`); see [discovery](../discovery/readme.md). TEA itself has no centralized authority that can police UUIDs across servers, so cross-server uniqueness is not enforceable and is not claimed.
-- The **object type** scopes uniqueness to a single object class (Product, Product Release, Component, Release, Collection, Artifact). A TEA server MUST guarantee that UUIDs are unique within `(authoritative domain, object type)`. UUIDs are not required to be unique across object types.
+- The **object type** scopes uniqueness to a single object class (Product, Product Release, Component, Component Release, Collection, Artifact). A TEA server shall guarantee that UUIDs are unique within `(authoritative domain, object type)`. Except as noted below, UUIDs are not required to be unique across object types.
```
ppkarwasz, line 14: Rather than listing all the object types and then adding an exception for Product/Component Release, could we list only those that need unique identifiers directly?

1. Products
2. Components
3. Collections
4. Artifacts
5. Distributions

A TEA server shall guarantee that UUIDs are unique within (authoritative domain, object type). UUIDs are not required to be unique across namespaces.

Every Product Release and every Component Release has exactly one Collection and shares its UUID.

doc/tea-uuid-scope.md
```diff
@@ -11,17 +11,19 @@ authoritative domain  +  object type  +  UUID
...
+**Exception: Product Release and Component Release.** A TEA Collection shall use the same UUID as its parent Product Release or Component Release. Within an authoritative domain, a Product Release and a Component Release shall not share a UUID. Consequently, Collections belonging to different parent releases have different UUIDs. Versions of the same Collection retain the same UUID. Except for Product Release and Component Release, objects of different types may share a UUID.
 
-Implementations MAY use generators that produce globally unique UUIDs (e.g. RFC 4122). Doing so neither violates nor strengthens this specification.
+The per-object-type scoping reflects how TEA servers are typically implemented: each object type lives in its own database table, where uniqueness is trivially enforced by a primary key. Cross-table uniqueness is not required in general (TEA has no operation that resolves a UUID without already knowing its object type), except that Product Release and Component Release UUIDs shall be disjoint as above, because those types share the Collection UUID namespace.
```
ppkarwasz, line 18: Doesn't this paragraph contain too many implementation details (which are left to the implementor)? AI agents often like to tell the story of their life…

taleodor:
@ppkarwasz If this is cosmetic (as I read it), I suggest we merge as is and then you can open small follow up to improve. This is currently blocking me on #313 which is a bigger change and needs to go in isolated.

Mehrn0ush:
Listing the independently identified entities and explaining that releases share their Collection’s UUID could simplify the model. However, “every
     release has exactly one Collection” alone does not explicitly prevent two releases from sharing that same Collection. A rewrite should preserve the
     relationship in both directions: each release has one logical Collection, each Collection belongs to exactly one release, and Collection versions
     retain its UUID. It should also distinguish object types from UUID namespaces. I would treat this as a considered rewrite rather than simply replacing
     the list.

> > The restriction is overly broad now. Can we take a step back and go for narrower fix from previous commit state?
> 
> We are talking about 128-bit numbers. If properly generated, all the TEA servers in the world can be merged into a single one, without any collision. Do we really need to have separate namespaces for different objects?



Agree. properly generated UUIDs make accidental collisions. The narrower question here is which UUID reuse the specification permits and what uniqueness the server is required to enforce. 
This PR preserves the existing per-type scope and prohibits only the release-type reuse that causes the Collection identity collision. I suggest considering a broader uniqueness requirement separately.

Mehrn0ush:
> > > The restriction is overly broad now. Can we take a step back and go for narrower fix from previous commit state?
> > 
> > 
> > We are talking about 128-bit numbers. If properly generated, all the TEA servers in the world can be merged into a single one, without any collision. Do we really need to have separate namespaces for different objects?
> 
> We had this conversation already. In practice collisions happen more frequently than in theory, for various reasons. The issue on the implementation side is that it's easy to enforce no collision rule within a single table, but more difficult across tables.

Understood. The current revision keeps the additional uniqueness requirement limited to Product Release and Component Release. It specifies the required result while leaving the enforcement mechanism to the implementation.

doc/tea-uuid-scope.md
Mehrn0ush: Agree. That seems a useful way to simplify the presentation. 
For clarity, this PR intentionally addresses the Collection identity collision only. The separate concern about different TEI authorities sharing an API base—and assigning the same UUID to different resources—is still open (As discussed in Slack and waiting for feedback). The proposed  (authoritative domain, object type) scope does not resolve that addressing ambiguity, right? so is it better to keep that decision separate from this PR?

Issue #312, opened 2026-09-19
Harmonise operationIds in openapi.yaml


ppkarwasz:
## Problem

`operationId` values in `spec/openapi.yaml` mix four different conventions. Since operationIds become method names in every generated client, this inconsistency is visible to all API consumers.

- The `Tea` prefix appears in 7 of 30 operations (`getTeaProductByUuid`, `queryTeaComponents`) and is absent from the rest (`getComponentReleaseById`, `getLatestArtifact`).
- The same kind of lookup is suffixed `ByUuid` for products and product releases but `ById` for components and component releases.
- Component release is the unmarked default for collection operations (`getLatestCollection`, `getCollectionsByReleaseId`, `getCollection`), while the product side is qualified (`getLatestCollectionForProductRelease`, `getCollectionsByProductReleaseId`, `getCollectionForProductRelease`). Nothing in the paths justifies the asymmetry.
- `discoveryByTei` is a noun rather than a verb and is inaccurate, since the endpoint also accepts `purl`.

## Proposal

Two candidate schemes are shown below. Either one is acceptable; the point is to pick one and apply it everywhere.

**Scheme A, parent-first.** `<verb><Parent><Child>[ByVersion]`. No `Tea` prefix, no `ById` suffix on single-item lookups. Verbs: `query` for top-level searches, `list` for parent-scoped listings, `get` for a single item, `download` for content. `Latest` goes before the noun and `ByVersion` after it, as the artifact operations already do. (`list` could be replaced by `get`, since the plural noun already carries cardinality; the only cost is that `getProductReleases` and `queryProductReleases` then read almost the same.)

**Scheme B, Spring Data style.** `<verb><Subject>By<Property>[And<Property>]`, as in Spring Data derived query methods: the subject names what is returned, the predicate after `By` names the path or query parameters used to select it. This is the convention the current `getReleasesByProductId` and `getCleByProductReleaseId` already follow. Cardinality is carried by the subject noun, not by the verb, so `get` is used throughout for reads and `download` for content.

| Path | Current | A: parent-first | B: Spring Data style |
|---|---|---|---|
| `/products` | `queryTeaProducts` | `queryProducts` | `getProductsByIdentifier` |
| `/product/{uuid}` | `getTeaProductByUuid` | `getProduct` | `getProductByUuid` |
| `/product/{uuid}/releases` | `getReleasesByProductId` | `listProductReleases` | `getReleasesByProductUuid` |
[...]

PR #334, opened 2026-09-21
refactor: harmonise operationIds and add operation summaries (#312)

taleodor:
Scheme A from #312, plus the summaries the scheme needs to be free of the published document.

operationIds now read `<verb><Parent><Child>[ByVersion]`: no `Tea` prefix, no `ById` / `ByUuid` on single-item lookups, `query` for top-level searches, `list` for parent-scoped listings, `get` for a single item, `download` for content, and the product and component sides named symmetrically. 21 of 28 change; the artifact and token operations already followed the scheme. `discoveryByTei` becomes `discover`, since the endpoint also accepts `purl`.

Every operation also gains a one-line `summary`. The file had none, and the Ecma publication pipeline renders `summary || operationId` as the clause heading, so until now the operationIds were the section titles of the standard: "6.3.4 downloadArtifactByVersion". With summaries the headings read as prose and the operationId is free to follow code convention. Rendered through the ECMA-xxx-TEA generator, all 28 clause headings are now sentences.

operationId is not part of the wire contract: paths, methods, parameters and payloads are unchanged, so a conforming server stays conforming. What changes is the method name in generated clients, which is why this belongs before 1.0 rather than after.

Verified with openapi-generator v7.24.0, the version pinned in CI: go, python, typescript and java-webclient all generate without errors and carry the new names (`queryProducts`, `getProduct`, `query_products`, ...). Spec valid; all 43 embedded examples still validate.


ppkarwasz (approved):
LGTM!

Mehrn0ush:
LGTM.

oej (approved):

PR #335, opened 2026-09-22
docs: require new revisions when external artifact URLs change (#330)

### LINKED ISSUE #330: Clarify how renewable artifact URLs interact with collection immutability and signature
## Summary

Artifact format `url` values may be renewed in place without a new artifact revision (#300), while collections embed those artifact objects and are described as immutable, optionally signed revisions. Renewing a URL either changes a collection’s serialized bytes (and any signature over them) or forces collection revisions solely for URL churn. The spec does not say which is intended.

Verified on `main` @ `86356888`.

## Current gaps

1. **In-place URL renewal is allowed on artifacts**  
   [`artifact-format.url`](https://github.com/CycloneDX/transparency-exchange-api/blob/86356888a7fda3af1da347ea8bc1f1f62ee018fe/spec/openapi.yaml#L1874-L1880): a server may replace an expired/renewed pre-signed URL in place; that is not a new artifact revision; content and checksums stay the same.

2. **Collections are immutable revisions and may be signed**  
   [tea-collection.md](https://github.com/CycloneDX/transparency-exchange-api/blob/86356888a7fda3af1da347ea8bc1f1f62ee018fe/tea-collection/tea-collection.md#L10-L24): each collection version is immutable; updates get a new version and may be signed as JSON.

3. **Collections embed full artifact objects**  
   A collection response includes nested `artifacts[].formats[].url`. Renewing a URL changes the collection JSON even when artifact content bytes and checksums are unchanged.

4. **Missing rules for**
   - whether URL-only changes require a new collection `version`
   - what historical `GET .../collection/{version}` must return after renewal
   - how signatures over a collection interact with transient URLs
   - whether clients should treat `url` as non-canonical for signing (e.g. strip/normalize before verify)

Without an explicit exception or renewal policy, implementers must choose between:

- mutating an “immutable” collection (and invalidating signatures), or
- minting collection revisions for every pre-signed URL refresh (noise for consumers).

## Proposed direction

Normatively state one of:

1. **URL-only refresh does not bump collection version**, and signatures (if any) MUST cover a defined canonical form that excludes transient `url` / `signatureUrl` fields; historical retrieval may return refreshed URLs for the same version; or
2. **Any change to the served collection document**—including URL renewal—**SHALL** produce a new collection version, with `updateReason` guidance for URL renewal; or
3. **Collections MUST NOT embed dereferenceable URLs** for signed/static collections; clients always use TEA download endpoints (aligns with parts of #321).

Also clarify interaction with artifact-level “renew in place” from #300 so the two layers do not contradict.

## Relation to other work

- Follow-up to #300 / leftover from #287 (refreshable non-content metadata).
- Related to #321 (drop direct download URLs), but this issue is needed even if #321 is deferred.

Mehrn0ush:
## Summary
- Close #330 with option 2: a change to published external `url` / `signatureUrl` shall create a new artifact revision, even when content bytes and checksums are unchanged.
- A collection publishes a new version only when it adopts that new artifact revision. Previously published collection versions remain unchanged.
- `url` and `signatureUrl` should be stable URLs. Servers should not publish pre-signed or otherwise expiring URLs in them. Expiring storage links are omitted, and the server answers from the download endpoints, with `302` where the bytes live elsewhere.
- Successful retrieval through the server's own download endpoints, including after redirects, yields unchanged artifact or signature bytes after HTTP transfer coding and content coding are removed. Changing only `Location` does not create a revision when those bytes stay the same.
- Checksum constraint: when `url` is present, `checksums` is required; a present `checksums` array shall be non-empty. Those checksums are the integrity statement for the external artifact bytes and do not cover detached signatures. A format with no `url` and no `checksums` remains valid.
- Precedence: when a later collection version mixes change kinds, `updateReason.type` is the first applicable value in this order: `VEX_UPDATED`, `ARTIFACT_REMOVED`, `ARTIFACT_ADDED`, `ARTIFACT_UPDATED`. The first collection version uses `INITIAL_RELEASE`. Adding a VEX that was absent from the preceding version counts as `ARTIFACT_ADDED`; updating an existing VEX counts as `VEX_UPDATED`. Clients monitoring VEX changes should inspect added artifacts as well as `VEX_UPDATED`.

## Test plan
- [ ] Confirm no remaining “replace URL in place / not a new revision” or “pre-signed URL in `url`” language
- [ ] A format with `url` and missing or empty `checksums` is rejected; a format with no `url` and no `checksums` is accepted; a present empty `checksums` array is rejected
- [ ] Skim `artifact-format.url` / `signatureUrl`, collection `version`, and `collection-update-reason-type`
- [ ] Skim `tea-collection` structure and the update-reason section

Closes #330


taleodor:
Thank you, good direction.

Four things; note the second asks for one small schema change.

**1. Pre-signed URLs in `url` become dead ends.** The text still contemplates them
("including renewing a pre-signed URL", "when the URL is pre-signed it can expire"). But
historical URLs shall be preserved, and the versioned download endpoint "is not required
to serve" a format that has a `url`, so every older revision carrying a pre-signed URL
becomes unretrievable once it expires. Each renewal also mints a new artifact revision and
a new collection version (a `VEX_UPDATED` for VEX). Separately, "Servers that need
short-lived … links without publishing an external URL shall omit `url`" is circular.
Suggest stating the #330 rationale directly, in both files:

> `url` and `signatureUrl` should be stable URLs. Servers should not publish pre-signed or
> otherwise expiring URLs in them; a server that serves content from expiring storage
> links shall omit them and answer from the download endpoints, with `302` where the bytes
> live elsewhere.

**2. Payload stability can't be required of external URLs; require checksums instead.**
"Successful retrieval … shall return unchanged payload bytes for as long as the resource
remains available" has no conformance target, and the TEA server doesn't control a third
party's bytes. Please keep the guarantee for the server's own download endpoints
(including `302`, where the server picks the `Location`), and for external URLs make the
published checksums the integrity statement, which then have to be present. In
`artifact-format.url`, replace that paragraph with:

> The TEA server's download endpoints shall return unchanged bytes for a fixed artifact
> UUID, version, and format, including when they answer with `302`. Changing only a
> download response's `Location`, while preserving those bytes, does not require a new
> artifact or collection version.
>
> The TEA server does not control the bytes behind an external `url`, so for every format
> that has a `url` the server shall publish at least one checksum: the checksums are the
> integrity statement for the revision. Content retrieved from `url` that does not match
> them is not the content of that revision.

and enforce it in the schema:

```yaml
    artifact-format:
      required:
        - mediaType
      dependentRequired:
        url:
          - checksums
      properties:
        checksums:
          type: array
          minItems: 1
```

I tested this: the spec still validates, all 43 examples pass (every existing format with
a `url` already has checksums, in the YAML and in `tea-collection.md`), and a format with a
`url` but missing or empty `checksums` is rejected. Formats without a `url` are unaffected.
`dependentRequired` is JSON Schema 2020-12, so it is available in OpenAPI 3.1.

Matching changes elsewhere:
- `artifact.version`: "Each published revision is immutable: the TEA server's download
  endpoints shall return unchanged payload bytes for a fixed artifact UUID, version, and
  format, including when they redirect."
- `tea-collection.md`, immutability paragraph: "TEA Artifacts are strictly **immutable**
  per revision: if the underlying document or any published field, including external
  `url` / `signatureUrl` values, changes, a new TEA Artifact revision shall be created.
  The TEA server's download endpoints shall return unchanged bytes for a fixed artifact
  UUID, version, and format, including when they redirect. The server does not control
  content behind an external URL; for those, the published checksums are the integrity
  statement for the revision."
- `tea-collection.md`, `url` bullet: replace the last sentence ("A published external URL
  can expire … does not authorize changing the payload bytes …") with "Servers shall
  publish at least one checksum for every format that has a `url`; content that does not
  match them is not the content of that revision."
- `tea-collection.md`, `checksums` bullet: mark it required when `url` is present, with at
  least one entry.

"Any published field" also settles whether, for example, adding a format needs a new
revision: the YAML already says every published revision is immutable, and the markdown
would then agree.

**3. Heavy duplication.** The URL-only classification sentence and the
collection-adoption rule each appear four times across the two files. Could one YAML copy
reference the other, as `signatureUrl` already does for `url`? Every copy is a place the
next edit can miss.

**4. Mixed change kinds.** `updateReason.type` is a single required value, and nothing
limits a collection version to one kind of change, so a version that adds an attestation
and revises the VEX, or one that re-points URLs on both VEX and non-VEX artifacts (now
more likely with this PR), has several valid values. That matters because
`VEX_UPDATED` is what clients alert on: a server that picks `ARTIFACT_ADDED` for the first
example is conforming, and a client watching for `VEX_UPDATED` misses the VEX change.
The gap predates this PR, but "Precedence … is not defined here" would ship twice in the
Ecma draft as normative text. Since the PR already edits both spots, suggest replacing the
sentence with:

> When a collection version other than the first contains changes of more than one kind,
> the server shall set `updateReason.type` to the first applicable value in this order:
> `VEX_UPDATED`, `ARTIFACT_REMOVED`, `ARTIFACT_ADDED`, `ARTIFACT_UPDATED`. `comment` can
> describe the other changes; a client that needs the complete set of changes should compare
> the version with the previous one.

Related, for the VEX signal to be reliable: "VEX artifact" isn't defined (the closest type
is `VULNERABILITIES`, which covers VDR and VEX), and `VEX_UPDATED` covers only revisions, so
publishing the first VEX for a release surfaces only as `ARTIFACT_ADDED`.

taleodor:
Thanks, this addresses the review: the stable-URL rule, the checksum constraint (I re-ran
it against this YAML: `url` without or with empty `checksums` is rejected, formats without
`url` pass, spec valid, 43/43 examples), stability scoped to the server's endpoints, and
the precedence order.

One open point on VEX. "`VEX_UPDATED` does not apply to every `VULNERABILITIES` artifact"
depends on a distinction the data model doesn't carry: VEX and VDR share the artifact type,
CycloneDX gives them the same media type, and one document can be both. So a client can't
tell whether a server chose `VEX_UPDATED` or `ARTIFACT_UPDATED` correctly. And with a new
VEX counted as `ARTIFACT_ADDED`, a version that removes an SBOM and adds a VEX reports
`ARTIFACT_REMOVED`, so putting `VEX_UPDATED` first in the order no longer guarantees VEX
changes surface. Suggest:

> `VEX_UPDATED` covers adding or revising an artifact of type `VULNERABILITIES`.

Nits:
- The download endpoint says expiring links "are not published in `url`"; `url` says
  "should not". Suggest "should not be published".
- The markdown `signatureUrl` bullet: "Successful signature retrieval … shall yield" has no
  subject; the YAML says "the server shall ensure".
- The byte-stability rule now appears in `artifact.version`, `url` and `signatureUrl`;
  keeping it in `artifact.version` and pointing to it from the other two would halve the
  repetition.

spec/openapi.yaml
```diff
@@ -1699,7 +1700,12 @@ components:
           type: integer
           description: |
             TEA Collection version, incremented each time its content changes.
-            Versions start with 1.
+            Versions start with 1. Content changes include replacing an embedded artifact
```
oej: and incremented by 1

spec/openapi.yaml
```diff
@@ -1907,13 +1947,32 @@ components:
...
+
             A TEA access token is sent only to the TEA server's own API base URL, never
-            to an external URL: external locations are either openly accessible,
-            pre-signed, or covered by credentials the client arranges separately. When
-            the URL is pre-signed it may expire; clients should not retain it beyond the
-            freshness lifetime of the response that carried it. A server may replace an
-            expired or renewed URL in place: that is not a new artifact revision, and the
-            content and checksums it points to are unchanged.
+            to an external URL. External locations are either openly accessible or covered
+            by credentials the client arranges separately.
```
oej: Change "covered by credentials"

tea-collection/tea-collection.md
```diff
@@ -53,7 +53,11 @@ The TEA Collection object has the following parts:
     See [TEA UUID Scope and Stability](../doc/tea-uuid-scope.md).
     When updating a collection, only the `version` is changed.
 - __version__: TEA Collection version, incremented each time its content changes.
-    Versions start with 1.
+    Versions start with 1. Content changes include replacing an embedded artifact with a
```
oej: increment by 1

taleodor:
Line 1648 of spec/openapi.yaml updateReason should become

```
              type: VEX_UPDATED
```

oej (approved):
Approved in meeting

PR #336, opened 2026-09-22
refactor: group paths, tags and token schemas by resource (#313)

### LINKED ISSUE #313: Group endpoints and tags by resource in openapi.yaml

taleodor:
## Summary

Closes #313. Two commits, deliberately separate because they need different kinds of review:

1. **`refactor: group paths, tags and token schemas by resource`** — pure reordering, verifiable mechanically.
2. **`refactor: tag /product/{uuid}/releases as TEA Product`** — one tag membership change, the only semantic edit in the branch.

## 1. Reordering

- **`paths`** ordered one block per resource in consumer-flow order — discovery, product, product release, component, component release, artifact, token — and within each block: list endpoint, single item, then sub-resources. The product and component sides now read the same way, the six collection endpoints no longer alternate between the two release types, and the artifact block keeps the `latest` operations together rather than interleaved with the versioned ones.
- **`tags`** reordered to match: Discovery, Product, Product Release, Component, Component Release, Artifact, CLE, Authentication.
- The three **`token-*` schemas** move from the top of `components.schemas`, ahead of the domain model, to the end beside the other authentication components.

Nothing is added, removed or edited.

## 2. The one retag

Every operation in the file is tagged by the resource in its path prefix, with `/cle` endpoints tagged `CLE` because lifecycle data is its own resource. I tabulated all 28 against that rule: **27 follow it, one does not.**

| path | tag | prefix implies |
|---|---|---|
| `/product/{uuid}/releases` | `TEA Product Release` | `TEA Product` |
| `/component/{uuid}/releases` | `TEA Component` | `TEA Component` ✓ |

Same shape, same role, opposite convention. Without this, the reordering leaves the "TEA Product Release" clause opening with "List the releases of a TEA Product", which reads as a bug in the standard rather than in the tag.

Tagging it `TEA Product` makes the four resource clauses symmetric:

```
TEA Product             Query TEA Products · Get a TEA Product · List the releases of a TEA Product
TEA Product Release     Query TEA Product Releases · Get a TEA Product Release with its latest collection · its collections
TEA Component           Query TEA Components · Get a TEA Component · List the releases of a TEA Component
TEA Component Release   Query TEA Component Releases · Get a TEA Component Release with its latest collection · its collections
```

**Impact:** the wire contract is unchanged — same path, method, parameters and payloads, so a conforming server stays conforming. Generated clients move `listProductReleases` from the Product Release API class to the Product API class, where `listComponentReleases` already sits on the component side. That is a source-compatibility change, but #312 renamed the same method in this release, so the call site is already being touched.

## Verification

Reordering PRs are not reviewable by reading the diff, so the first commit was checked mechanically instead:

| Check | Result |
|---|---|
| Parsed documents identical once the `tags` **list** is sorted | ✓ — a recursive deep-diff reports no residual difference, i.e. tag order is the only semantic change in commit 1 |
| Path-key, operationId and component-key sets, all five `components` sections | unchanged |
| All 28 operation bodies, canonically dumped | byte-identical |
| Tag membership per operation, commit 1 | unchanged |
| Tag membership per operation, commit 2 | exactly one change: `listProductReleases` |
| Spec + embedded examples | valid, 43/43 |
| openapi-generator v7.24.0 (the pinned CI version) — go, python, typescript, java-webclient | generate without errors; same eight API classes and 39 models; `ListProductReleases` now in the Product API class, mirroring `ListComponentReleases` |

## Note for the published standard

The `tags` order does not currently reach the Ecma document: the ECMA-xxx-TEA generator sorts tag sections alphabetically, so the standard would still open with CLE regardless of what this file declares. A one-function fix on that side makes it follow the declared order — I have it ready and will open it against `Ecma-TC54/ECMA-xxx-TEA` alongside this. Operation order *within* a section already follows `paths`, so that half of #313 carries through today.

The data model section stays alphabetical there: it is a catalogue looked up by schema name rather than a narrative sequence, so the `token-*` move in this PR is for readers of the YAML.


Mehrn0ush:
LGTM

ppkarwasz:
Looks good to me, with two _nitpicks_:

- What is the appropriate tag for `/product/{uuid}/releases` and `/component/{uuid}/releases`? "TEA Product Release / TEA Component Release" has the advantage that all the endpoints with the same tag return data from the same database table.
- Same question about `/productRelease/{uuid}/collection*` and `/componentRelease/{uuid}/collection*`: maybe a new "TEA Collection" tag would be more appropriate? CLE endpoints have their own tag.

PR #337, opened 2026-09-22
refactor: tag operations by the resource they return

ppkarwasz:
Follow-up to #336, which grouped the paths and tags by resource and, in its second commit, retagged `/product/{uuid}/releases` as `TEA Product` so that every operation is tagged by its path prefix. This PR applies the other rule I suggested in the review there: tag every operation by the resource it returns, so that all operations sharing a tag return the same object type, the way the `CLE` endpoints already do.

Two changes, no path, parameter or payload touched:

- `/product/{uuid}/releases` and `/component/{uuid}/releases` are tagged `TEA Product Release` and `TEA Component Release`. They list releases, and a client looking for "how do I get the releases" finds them next to `/productReleases` and `/componentReleases`.
- A new `TEA Collection` tag, placed between `TEA Component Release` and `TEA Artifact` in the `tags` list, covers the six `/collection*` endpoints: `/productRelease/{uuid}/collections`, `/collection/latest`, `/collection/{collectionVersion}` and their component-release twins. Collections are their own object type with their own schema and pagination, and they no longer split across the two release tags.

The resulting tag membership:

| Tag | Operations |
|---|---|
| TEA Product | `/products`, `/product/{uuid}` |
| TEA Product Release | `/product/{uuid}/releases`, `/productReleases`, `/productRelease/{uuid}` |
| TEA Component | `/components`, `/component/{uuid}` |
| TEA Component Release | `/component/{uuid}/releases`, `/componentReleases`, `/componentRelease/{uuid}` |
| TEA Collection | the six `/collection*` endpoints |
| CLE | the four `/cle` endpoints |
| TEA Artifact, TEA Discovery, TEA Authentication | unchanged |

Impact on generated clients: a new `TeaCollectionApi` class receives the six collection operations, and the two release listings sit in `TeaProductReleaseApi` and `TeaComponentReleaseApi`. The wire contract is unchanged. Verified with openapi-generator 7.26.0: the spec validates and the Java webclient generates nine API classes.

The order of `paths` from #336 is kept as is, so the YAML still reads one block per path prefix while the tags group by return type. In the Ecma document the tags decide section membership, so the "TEA Product Release" section will open with the listing of a product's releases; I think that is the right place for it.


Mehrn0ush:
LGTM

taleodor (approved):

PR #338, opened 2026-09-23
Light editorial changes for publication

gesa:


discovery/readme.md
```diff
@@ -17,17 +17,17 @@
...
+- how discovery is initiated;
+- how discovery documents are retrieved;
+- how API endpoints are obtained.
 
 TEA separates:
 
-- **identity** → TEI  
-- **location** → discovery  
-- **data retrieval** → API  
+- **location** → discovery
```
oej, line 28: Location and discovery is based on TEI so TEI should come first

doc/tea-requirements.md
```diff
@@ -56,16 +57,20 @@ urn:cdx:serialNumber/version
...
 
 The API needs to provide support for update checks, i.e. to check if a document is
-updated without downloading. (possibly etag or HEAD method or similar)
-Authn/Authz MUST be supported
+updated without downloading.
+
+> Note: Possible approaches include etag or HEAD method.
 
-## Artefact Publishing
+Authn/Authz MUST be supported.
```
oej, line 64: Actually it is optional.

doc/tea-requirements.md
```diff
@@ -56,16 +57,20 @@ urn:cdx:serialNumber/version
...
+
+## Artefact publishing
 
 The API MUST provide a way to publish an artefact, either standalone or to a collection. 
 The detection of duplicate artefacts with the same identity MUST be handled and prevented.
-Authn/Authz MUST be supported
 
-## Artefact Versioning
+Authn/Authz MUST be supported.
```
oej, line 71: Optional, but supported by the API

(For Open Source, auth does not make much sense and they may want to create a very simple implementation based on existing file systems - which is why we support the external URLs).

oej (changes requested):
Some changes needed around auth being required, which is not the fact.

oej:
Thank you for your contribution!

oej:
In order to proceed, we're replacing this PR with another one we can modify.

PR #350, opened 2026-09-23
docs: separate artifact checksums from HTTP representation validators

### LINKED ISSUE #349: Reconcile HTTP content coding with checksums and ETags
# Summary
As discussed in Slack: Gzip and identity responses for the same artifact are different HTTP representations. Metadata checksums shall cover the artifact bytes before content coding. Strong ETags and Repr-Digest shall follow the selected representation, Vary shall include Accept-Encoding when encoding is negotiated, and immutable applies only when those content-coded bytes stay unchanged for the freshness lifetime. Changing content coding alone does not create a new artifact or collection revision.

Mehrn0ush:
## Summary
Rebased onto `main` after #335. The #335 text is kept as merged, and only the HTTP rules are added on top.

- Published checksums are calculated over the artifact bytes before HTTP content coding. Clients remove transfer and content coding first and do not otherwise transform the bytes. A `.gz` artifact stays compressed, and servers shall not declare an artifact's own compression as HTTP content coding.
- Each selected HTTP representation, including its content coding, has its own strong `ETag`. Intermediaries that weaken the tag when compressing on the fly do not meet this; servers can use precompressed variants or their own strong tags.
- `Repr-Digest` covers the content-coded representation and does not replace the published checksums.
- `immutable` is allowed only when the content-coded bytes stay unchanged for the freshness lifetime.
- New `Content-Encoding` header (omitted when no coding is applied). `artifact-vary` covers `Accept` and `Accept-Encoding`, including absent headers and `*/*`. `304` repeats `ETag` and `Vary`.
- `Content-Location` identifies the revision and format; content coding can still be negotiated.
- Changing only HTTP content coding does not create an artifact or collection revision.

Closes #349

## Test plan
- [ ] Two different gzip encodings of the same artifact share a published checksum and have different strong ETags and Repr-Digests
- [ ] `If-None-Match` for one encoding does not match the other; `304` repeats that representation's `ETag` and `Vary`
- [ ] Repeat those checks for signature downloads
- [ ] A format with `url` and missing or empty `checksums` is still rejected


taleodor:
Thanks, the HTTP content here is right and fills a real gap. It needs a rebase onto the merged #335, and that rebase is where it can go wrong.

### Rebase

A trial merge with `main` gives eight conflict hunks: five in `spec/openapi.yaml` and three in `tea-collection/tea-collection.md`. The largest is `/artifact/{uuid}/{artifactVersion}/download`, which `main` has moved. This branch's side of those hunks still carries four phrasings that #335 removed after review:

- "for as long as the resource remains available": the payload promise for external URLs, which #335 replaced with checksums as the integrity statement;
- "Servers that need short-lived … shall omit `url` / `signatureUrl`": the circular sentence;
- "pre-signed, or covered by credentials";
- "pre-signed URL, revision" in the `signatureUrl` description.

A safe resolution keeps `main`'s text in every hunk, then re-applies only what is new here:

- **Checksums.** They are computed over the artifact bytes before HTTP content coding. Clients remove transfer and content coding first and do not otherwise transform or canonicalize the bytes, and a `.gz` artifact stays compressed. Keep #335's rule that `checksums` is required whenever `url` is present and has at least one entry.
- **ETag.** Each selected HTTP representation gets its own strong `ETag`, content coding included.
- **Repr-Digest.** It covers the content-coded representation data and does not replace the metadata checksums.
- **`immutable`.** It is allowed only when the content-coded bytes stay unchanged for the whole freshness lifetime.
- **Headers.** The new `Content-Encoding` header, the `Vary` rules for `Accept` and `Accept-Encoding`, and `304` repeating `ETag` and `Vary`.
- **`Content-Location`.** It identifies the revision and format, while content coding can still be negotiated.
- **Revision rule.** "Changing only HTTP content coding" can go into the `Location` sentence on `artifact.version` in `main`, which already says the bytes compare after content coding is removed.

### Substance

This matches RFC 9110 and RFC 9530: gzip and identity responses get distinct strong entity tags, `Vary` is sent on the uncompressed default response too, `304` repeats `ETag` and `Vary`, `Repr-Digest` covers the representation as sent including content coding, and `identity` is never named in `Content-Encoding`.

### Suggestions

1. **Make the `.gz` rule bind the server.** At the moment it only tells the client what to do. Suggest stating that an artifact's own compression shall not be declared as HTTP content coding. Otherwise clients decode it automatically and the checksum fails, which is a common object-storage misconfiguration.
2. **Rename `artifact-vary-accept`.** It now covers `Accept-Encoding` as well; `artifact-vary` would fit.
3. **One name for the checksums.** This PR says "metadata checksums", while the text #335 merged says "published checksums".
4. **Strong entity tags and on-the-fly compression.** Requiring a strong `ETag` on content-coded responses rules out proxies that weaken the tag when they compress on the fly, nginx among them. Servers will need precompressed variants or their own strong tags. A sentence saying so would save implementers the surprise.

Mehrn0ush:
Thanks @taleodor . I rebased onto main and kept main's text in every conflict, so the four phrasings are gone, then reapplied only the HTTP rules. 
The four suggestions are in: servers shall not declare an artifact's own compression as content coding, artifact-vary-accept is now artifact-vary, the text says "published checksums" throughout, and artifact-etag notes the on-the-fly compression pitfall with precompressed variants or per-coding strong tags as the way out.

spec/openapi.yaml
```diff
@@ -2793,9 +2866,9 @@ components:
         The selected representation is available at another location, for example object
         storage addressed by a pre-signed URL. The client follows the `Location` header.
 
-        Changing only this response's `Location`, while preserving the bytes for the
-        selected artifact UUID, version, and format, does not require a new artifact or
-        collection version. Publishing a different external `url` or `signatureUrl` on
+        Changing only this response's `Location`, while preserving the artifact or
+        signature bytes for the selected UUID, version, and format after HTTP content
+        coding is removed, does not require a new artifact or collection version. Publishing a different external `url` or `signatureUrl` on
```
taleodor: pls fix line length here

tea-collection/tea-collection.md
```diff
@@ -144,22 +145,29 @@ A TEA Artifact object contains the following fields:
     When `url` is present, these checksums are the integrity statement for the revision.
     Content that does not match them is not the content of that revision.
 
+    Published checksums shall be calculated over the artifact format's bytes before HTTP
+    content coding is applied. Clients verifying them shall first remove any HTTP transfer
+    coding and content coding, and shall not otherwise transform or canonicalize the
+    artifact bytes. An artifact that is itself compressed, such as a `.gz` file, stays in
+    that form: servers shall not declare the artifact's own compression as HTTP content
```
taleodor: let's change to "...stays in that form. The TEA servers shall not declare..."

taleodor (changes requested):

Mehrn0ush:
Done

taleodor (approved):

PR #361, opened 2026-09-23
Remove old BOM exchange file

oej:


ppkarwasz:
This PR only removes the `NOTICE` file. The old BOM exchange specification is in `archive` and can be removed.

PR #364, opened 2026-09-23
docs: scope UUID uniqueness to the API base (#363)

### LINKED ISSUE #363: UUID uniqueness must match how endpoints address objects
## Problem
`doc/tea-uuid-scope.md` requires UUID uniqueness only within `(authoritative domain, object type)`. UUID paths such as `/productRelease/{uuid}` do not carry the authoritative domain. If two TEI authorities share one API base and assign the same product-release UUID to different releases, both resolve to the same path.
## Proposal
- Require uniqueness per object type within an addressable API base.
- Authorities whose UUIDs would collide shall be served from separate API bases.
- Recommend that implementations create UUIDs with a cryptographically secure random number generator (RFC 9562 version 4, or version 7 with its random bits drawn that way), so independently operated servers are very unlikely to mint the same UUID. This is a recommendation. The protocol still cannot assign UUIDs across servers, and a server shall reject a UUID already in use at its own API base.
Related to #217 and #245.

Mehrn0ush:
## Summary
- UUID retrieval uniqueness is `(API base, object type, UUID)`, because paths such as `/productRelease/{uuid}` do not carry the TEI authoritative domain.
- Authorities that would assign the same UUID to different objects of the same type shall use separate API bases.
- Implementations should create UUIDs with a cryptographically secure random number generator. A server shall still refuse a UUID already in use at its API base.
## Test plan
-  Read `doc/tea-uuid-scope.md` and confirm the Product Release / Component Release exception still shares the Collection UUID
-  Confirm `spec/openapi.yaml` and `tea-collection/tea-collection.md` say "API base" for that exception
Closes #363


taleodor (changes requested):
The reasoning is right: a path such as `/productRelease/{uuid}` carries only the API base and the UUID, so uniqueness has to be scoped to what a client can address. Two things before merging.

### Must fix: two descriptions still use the old scope

`spec/openapi.yaml` still scopes the release-type rule to the authoritative domain in the two release UUID descriptions. This PR changes the same sentence in the Collection `uuid` description and in `tea-collection/tea-collection.md`, so after merging the spec would state the rule both ways.

Line 1181, TEA Product Release `uuid`:

```yaml
            A unique identifier for the TEA Product Release. Within an API base it
            shall not also identify a Component Release; see `doc/tea-uuid-scope.md`.
```

Line 1357, TEA Component Release `uuid`:

```yaml
            A unique identifier of the TEA Component Release. Within an API base it
            shall not also identify a Product Release; see `doc/tea-uuid-scope.md`.
```

### Optional: single spaces in the tuple

Line 14 of `doc/tea-uuid-scope.md` puts double spaces around the plus signs. This document is a normative chapter in the Ecma draft, and those double spaces produce four ecmarkup lint warnings there. Since this PR rewrites the line anyway:

```text
API base + object type + UUID
```

Mehrn0ush:
Thank you. 
Done.

PR #365, opened 2026-09-23
Add RFC reference for Well-known URIs

oej:
In answer to @gesa question in Slack


taleodor:
RFC 8615 reference. It adds RFC 8615 to the references, but the text above still cites "Section 1.1 of [RFC5785]". RFC 5785 isn't in that list, and RFC 8615 obsoletes it. The citation should change to RFC 8615 in the same PR.

oej:
Missed this, will open an issue and fix in another PR. Sorry.

Issue #324, opened 2026-09-20
Drop `sortField` and `sortOrder`: they block static hosting and buy consumers little


ppkarwasz:
Static hosting is what most open source projects have, and it is enough for most of the TEA specification.

Apart from the query endpoints (`/discovery`, `/products`, `/productReleases`, `/components`, `/componentReleases`) and `/token`, every endpoint returns a fixed JSON document per UUID. Those documents can be generated at release time and served by any web server, GitHub Pages included. I expect many open source projects to stop there. Running a compliant TEA server is a burden they have no reason to carry, so initial adoption would otherwise be limited to the companies that must place a product on the EU market and not reach their upstream.

The query endpoints can be pre-generated too, with a rewrite rule that turns

```
/componentReleases?idType=PURL&idValue=pkg%3Amaven%2Forg.apache.logging.log4j%2Flog4j-core
```

into a path such as `/componentReleases/PURL/pkg%3Amaven%2Forg.apache.logging.log4j%2Flog4j-core.json`. Pagination is not an obstacle: most projects have few enough releases and components to fit in one page, so the file holds everything and says `hasNext: false`. Such a server is not compliant to the letter, but it is compliant in practice because no client verifies the page size it asked for.

`sortField` and `sortOrder` are where this stops being reasonable. Pre-generating one answer per identifier query is a maintenance cost a project can accept. Pre-generating one answer per identifier query *and* per sort field *and* per direction is not: on the release endpoints that is six copies of every result, all containing the same rows. A static publisher will generate one order and ignore the parameters, and the spec gives it no way to say so, so a client that asked for `sortOrder=desc` silently gets the file's order.

## What the parameters cost and give

- For products, components and collections `sortField` has exactly one allowed value, so the parameter selects nothing. Its description for collections admits the tie-breaker "is redundant unless additional collection sort fields are added later".
- For releases, sorting by `version` is "the server's documented string collation; semantic-version precedence is not implied". Two servers may order the same releases differently and both are correct, so a client cannot rely on the result anyway.
- The tie-breaker and continuation rules from #253 exist to keep sorting stable across pages. An opaque `pageToken` already guarantees that on its own.
- A consumer that wants a different order sorts the page it received.

## Proposal

[...]

PR #368, opened 2026-09-23
feat: let a static file server be a conforming TEA server

ppkarwasz:
Follow-up to #324, in particular https://github.com/CycloneDX/transparency-exchange-api/issues/324#issuecomment-5758931237. Many open source publishers might serve TEA as pre-generated JSON files behind a plain web server. Three things in the spec make that impossible today; this PR removes them.

**1. Discovery takes path parameters.** `/discovery?tei=…` becomes `/discovery/{domain}/{type}/{identifier}`: the request path is the TEI itself, with `tei://` replaced by `<api base>/v<version>/discovery/`.

**2. The search endpoints are optional.** `/products`, `/productReleases`, `/components` and `/componentReleases` are not on the TEI-based consumer flow and are the only endpoints that need code. A server that does not implement them answers `404`, with or without a TEA error body. Since an implemented search endpoint never answers `404` (no match is `200` with an empty list), a `404` on these paths is unambiguous on its own.

**3. The TEA error body on `404` is optional.** Requiring a specific body to say "not implemented" means implementing something for an endpoint that is not implemented, and a static file server cannot attach one to its own `404`. The `404-object-by-id-not-found` description now says the status alone carries the meaning and a server may add the body to say which case applies.

**One TEI definition change, please review explicitly.** The `hash` TEI type was the only one whose identifier spanned two path segments (`hash/<hashtype>/<hash>`). To keep every TEI as `<domain>/<type>/<identifier>`, it becomes single-segment: `tei://<domain-name>/hash/<hashtype>:<hex>`. The colon is an RFC 3986 `pchar` and never needs escaping.

Docs updated accordingly: `discovery/readme.md` (three-step URL construction, examples, "Discovery by PURL", hash type, `404` and failover wording) and `api-flow/consumer.md` (diagram label and a note that search is optional). Mentions of `/discovery` as the endpoint's name in untouched paragraphs are left as they are.

Verified with openapi-generator v7.24.0 (the CI version): the spec validates, and the go, python, typescript and java-webclient clients generate. `TeaDiscoveryApi.discoverByTei(domain, type, identifier)` and `discoverByPurl(identifier)` replace `discover(tei, purl)`; the four `query*` methods are unchanged apart from the documented `404`. No `tei=` or `purl=` remains outside `archive/`, and every anchor in `discovery/readme.md` still resolves.

Out of scope, tracked separately: `sortField`/`sortOrder` (#324 itself) and a capability flag in `.well-known/tea`.


taleodor:
This is a major rewrite of the spec. We should discuss path forward past 1.0.0, not something we could entertain now.

oej:
I think we have to put this in the backlog as we are stressed and out of time to make major changes. Sorry.

discovery/readme.md
```diff
@@ -116,7 +116,7 @@ Reminder: the `unique-identifer` component of the TEI needs only be unique withi
 #### PURL - Package URL
 
 Where the `unique-identifier` is a PURL in it's canonical string form.
-A PURL identifier is encoded using BASE64URL.
+A PURL identifier is encoded using BASE64URL (RFC 4648 section 5), padding omitted.
```
oej: This needs to go in a separate PR

discovery/readme.md
```diff
@@ -140,15 +140,17 @@ Where the `unique-identifier` is a Hash. Supports the following hash types:
...
-````
+tei://<domain-name>/hash/<hashtype>:<hash>
+```
 
 Example:
 
 ```text
-tei://cyclonedx.org/hash/SHA256/fd44efd601f651c8865acf0dfeacb0df19a2b50ec69ead0262096fd2f67197b9
-
+tei://cyclonedx.org/hash/SHA256:fd44efd601f651c8865acf0dfeacb0df19a2b50ec69ead0262096fd2f67197b9
```
oej: This really needs to be in the same separate PR

ppkarwasz:
@taleodor,                                                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                                                                                 
I opened this PR as a companion to #324: a concrete example of what the API would need to change for a static web server to serve it.                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                                 
I realise it is not a priority for 1.0.0; I wanted it on record for future work.                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                 
Whether this counts as a major or minor change to the core API is a matter of opinion. A couple of months ago `/discovery` did not exist at all. Either way, it needs more discussion than we have time for now.

discovery/readme.md
ppkarwasz: Split out as #370 (unpadded Base64URL for the `purl` TEI).

discovery/readme.md
ppkarwasz: Split out as #369 (single-segment `hash` TEI). #370 covers the PURL encoding; the two do not overlap, so they can be reviewed independently.

PR #371, opened 2026-09-23
docs: define how TEIs are compared

ppkarwasz:
The discovery document defines the TEI syntax and its types but never says when two TEIs are the same identifier. This PR adds a "Comparing TEIs" section to `discovery/readme.md`, between "TEI types" and "TEI resolution using DNS".

The rule: two TEIs are equal if and only if they are the same sequence of characters. The comparison is case-sensitive and applies to the TEI as written; percent-escapes are neither added nor removed, Base64URL identifiers are not decoded, and the domain name is neither resolved nor normalized. Five example TEIs show spellings that are therefore distinct (domain case, UUID case, padded versus unpadded Base64URL).

Because a strict string rule would let two spellings of one identifier diverge, vendors shall publish a TEI in its canonical form (lowercase domain, the identifier exactly as its type defines it, no unnecessary percent-escapes) and clients shall use a TEI exactly as received. The section closes by separating identity from resolution, where the domain name is a DNS name and therefore case-insensitive.

This is the comparison model used for XML namespace names, stated in TEA's own terms so that no additional normative reference is needed. The unpadded Base64URL example relies on #370.


taleodor:
String equality is the right model, and separating identity from resolution is clear. A few gaps in the canonical form:

1. UUID case: the example treats an uppercase UUID as a different TEI, but the UUID type doesn't define case. Suggest adding "lowercase for a UUID (RFC 9562)".
2. Scheme: under case-sensitive comparison `TEI://` differs from `tei://`; the canonical form should include "the scheme `tei` in lowercase".
3. Internationalised domains: "lowercase" doesn't pick between A-labels and U-labels. Suggest "lowercase ASCII, using A-labels for internationalised names".
4. Transport: `/discovery` carries the TEI in the `tei` query parameter, percent-encoded. Suggest stating that comparison applies to the TEI after that encoding is removed, and that a server matching `tei` uses this rule.

Also, #315 currently says the type is "case insensitive and lowercase", which contradicts this; whichever merges second should reduce it to "lowercase".

discovery/readme.md
```diff
@@ -243,6 +244,37 @@ releases returned per TEI. Preferable situation is to have a single product rele
 per TEI. When multiple releases are returned, clients shall treat array order as
 priority (first entry highest).
 
+### Comparing TEIs
+
+A TEI is an identifier, not a locator.
+Two TEIs are equal if and only if they are the same sequence of characters.
+The comparison is case-sensitive and applies to the TEI as written:
+percent-escapes are neither added nor removed,
```
oej: Bullet list here

discovery/readme.md
```diff
@@ -243,6 +244,37 @@ releases returned per TEI. Preferable situation is to have a single product rele
...
+Base64URL-encoded identifiers are not decoded,
+and the domain name is neither resolved nor normalized.
```
oej: It really is but not for comparison. We may need to rewrite this

discovery/readme.md
```diff
@@ -243,6 +244,37 @@ releases returned per TEI. Preferable situation is to have a single product rele
...
+
+The following TEIs are therefore all distinct:
+
+```text
+tei://example.com/uuid/62f2cf92-ae88-11f1-a698-1a52914d44b2
+tei://Example.com/uuid/62f2cf92-ae88-11f1-a698-1a52914d44b2
```
oej: I don't think we should say that "example.com" and "Example.com" is different since we do resolve and they are the same domain.

discovery/readme.md
```diff
@@ -243,6 +244,37 @@ releases returned per TEI. Preferable situation is to have a single product rele
...
+tei://example.com/uuid/62F2CF92-AE88-11F1-A698-1A52914D44B2
+tei://example.com/purl/cGtnOnB5cGkvY3ljbG9uZWR4LXB5dGhvbi1saWI
+tei://example.com/purl/cGtnOnB5cGkvY3ljbG9uZWR4LXB5dGhvbi1saWI=
+```
+
+So that one identifier has one spelling,
+a vendor shall publish a TEI in its canonical form:
+the domain name in lowercase,
```
oej: Bullet list

oej:
Good work, this text is needed.

discovery/readme.md
ppkarwasz: This part reflects [§2.3 of Namespaces in XML 1.0](https://www.w3.org/TR/xml-names/#NSNameComparison). I don't know what was the W3C motivation for comparing a namespace, but it makes sense:

- Users should store TEIs **as opaque identifiers** without trying to be smart about them.

Otherwise we end up with the same situation as with RFC 822 addresses, where “smart” users take the liberty to capitalize and de-capitalize even the **local-part** of the e-mail address, even if it is case-sensitive.

discovery/readme.md
ppkarwasz: Fixed in https://github.com/CycloneDX/transparency-exchange-api/pull/371/commits/401118fd3c23af201323b2d69b990443d4c0a3de

discovery/readme.md
ppkarwasz: Fixed in https://github.com/CycloneDX/transparency-exchange-api/pull/371/commits/401118fd3c23af201323b2d69b990443d4c0a3de

ppkarwasz:
Thanks, all four are in 4ed3dfd: the canonical form now lists the scheme `tei` in lowercase, the domain in lowercase ASCII with A-labels for internationalised names, the type in lowercase, and lowercase for a UUID (the UUID type section now cites RFC 9562); a new paragraph says the comparison applies after the transport percent-encoding is removed and that a server matching a received TEI applies the rule.

On #315: "the type in lowercase" is the wording here, so whichever merges second reduces the other to that.

One consequence for the Ecma document: RFC 9562 is not in its normative references yet, although `openapi.yaml` already relies on it for the `uuid` schema. It needs to be added there *even* if the TEI type definitions are moved out of the specification into a registry, since the `uuid` schema stays.

oej:
I am happy with this even though I mumble about case-sensitive domain names... As long as we have a recommendation that when *publishing* a TEI, the domain part should always be lower case. If we can avoid the whole mess with smörgåstårta.se and other international domain names I'm happy.

taleodor:
1. The example list reads as if `tei://Example.com/…` were a second valid identifier. It
   isn't; the canonical form forbids publishing it. Suggest adding after the list: "Only
   the first and fourth are in canonical form; a vendor does not publish the others."
2. TEIs are meant to be printed on invoices, labels and "about" boxes and typed by people,
   and RFC 9562 treats UUIDs as case-insensitive on input. "A client shall use a TEI
   exactly as received and shall not rewrite it" forbids even lowercasing a UUID someone
   typed in uppercase. Suggest: "A client that accepts a TEI entered by a person may
   convert its scheme, domain name, UUID and hash digits to lowercase; otherwise a
   client shall use a TEI exactly as received." That covers oej's resolution concern at
   the one place it matters, without making comparison normalise.

taleodor (approved):

